PR #32, opened 2021-12-14
ScrollProgress performance improvement

### LINKED ISSUE #31: [oui/ScrollProgress] Calls getBoundingClientRect in scroll handler
https://github.com/noaignite/oui/blob/ef1cee994242187d06be4808d9a8196b4bc6df65/packages/oui/src/ScrollProgress/ScrollProgress.js#L22

Would it be possible to calculate scroll progress without continuously calling getBoundingClientRect? For example, if saving the bounding rect and window size only when resizing, it should be possible to calculate scroll progress without calling this on every scroll event.

I think this could be a good performance boost, considering getBoundingClientRect is one of the things that usually cause layout thrashing, and this component is often used specifically for re-rendering styles based on layout.

https://gist.github.com/paulirish/5d52fb081b3570c81e3a

alexanderflink:
This makes sure getBoundingClientRect is only called on window resize since it's an expensive
function

fix #31

### DIFF
--- a/packages/oui/src/ScrollProgress/ScrollProgress.js
+++ b/packages/oui/src/ScrollProgress/ScrollProgress.js
@@ -4,9 +4,15 @@ import { setRef } from '@mui/material/utils'
 import { clamp } from '@noaignite/utils'
 import InView from '../InView'
 
-export function calculateVerticalProgress(bounds, topOffset = 0, bottomOffset = topOffset) {
+export function calculateVerticalProgress(
+  bounds,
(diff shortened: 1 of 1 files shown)


alexanderflink:
Need to attach a performance benchmark and refactor to make the code a bit more readable

maeertin:
Closing this PR as we're dropping the oui package.

maeertin:
Closing this PR as we're dropping the oui package.

PR #34, opened 2022-01-12
Feature/accelerator merge

alexanderflink:
Massive PR here! You can ignore everything under `packages/react-centra-checkout/docs/`. I included it in version control so we can use github pages or similar to easily host documentation.


scripts/copy-files.js
```diff
@@ -40,10 +40,10 @@ async function createPackageFile() {
     ...packageDataOther,
     private: false,
     ...(packageDataOther.main && {
-      main: './index.js',
+      main: './node/index.js',
```
alexanderflink: @maeertin I changed this so that `main` points to the node (commonjs) version of the packages which lives under `node` and `module` points to the esm package which lives in the root. I believe this is how MUI does it and I've seen it in a few other places.

I need to update the PR to make the other packages follow this convention as well. What do you think?

PR #35, opened 2022-01-13
feat!: add react-centra-checkout package and update build setup

alexanderflink:
Redid the PR to not included generated docs.


alexanderflink:
@maeertin Moved the centra types into a separate package with separate generated docs

maeertin:
That's awesome, think it makes much more sense now!

babel.config.js
```diff
@@ -13,38 +14,26 @@ const useESModules = ['esm'].includes(process.env.BABEL_ENV)
...
 ]
 
 const plugins = [
   'babel-plugin-optimize-clsx',
-  // Need the following 3 proposals for all targets in .browserslistrc.
-  // With our usage the transpiled loose mode is equivalent to spec mode.
-  ['@babel/plugin-proposal-class-properties', { loose: true }],
-  ['@babel/plugin-proposal-private-methods', { loose: true }],
-  ['@babel/plugin-proposal-private-property-in-object', { loose: true }],
-  ['@babel/plugin-proposal-object-rest-spread', { loose: true }],
```
maeertin, line 39: Are these not needed @alexanderflink? They seem to still live in package.json :)

package.json
```diff
@@ -44,12 +44,17 @@
...
     "@commitlint/cli": "^15.0.0",
     "@commitlint/config-conventional": "^15.0.0",
     "@commitlint/cz-commitlint": "^15.0.0",
     "@testing-library/jest-dom": "^5.14.1",
     "@testing-library/react": "^12.1.0",
     "@testing-library/user-event": "^13.2.1",
+    "@tsconfig/recommended": "^1.0.1",
+    "@types/react": "^17.0.35",
+    "@typescript-eslint/eslint-plugin": "^5.4.0",
+    "@typescript-eslint/parser": "^5.4.0",
```
maeertin, line 57: Not sure if these `@typescript-eslint/parser` and similar packages need to live in the root package.json as they are included via `@noaignite/eslint-config` @alexanderflink

packages/react-centra-checkout/src/Context/index.tsx
```diff
@@ -0,0 +1,317 @@
...
+  },
+  paymentMethods: [],
+  selection: {
+    address: {},
+    discounts: {},
+    items: [],
+    totals: {},
+  },
+  shippingMethods: [],
+}
```
maeertin, line 55: I think these should have a more descriptive name like `SELECTION_INITIAL_VALUE` or so and also exported for certain occasions @alexanderflink

packages/react-centra-checkout/src/index.ts
```diff
@@ -0,0 +1,7 @@
+/* eslint-disable import/export */
+
+export { default as ApiClient } from './ApiClient'
+export * from './ApiClient'
+
+export { default } from './Context'
```
maeertin, line 6: Wondering if it makes more sense or not to have all exports being non default exports here 🤔 @alexanderflink Thoughts?

babel.config.js
alexanderflink: I removed them when updating the babel config since I figured they aren't needed for the browser targets we have. Not entirely sure how to test this though. I'll also remove them from package.json!

packages/react-centra-checkout/src/index.ts
alexanderflink: I think it's kinda nice to have default exports for each of them, don't we also have an eslint rule which requires a default export?

babel.config.js
maeertin: Awesome! Makes sense to remove them from package.json if they are not used in babel.config. Make sure to remove them though since I can still find them under `resolutions` in package.json ;) @alexanderflink

packages/react-centra-checkout/src/index.ts
alexanderflink: Ahh, now I see what you mean. Totally makes sense. Will change that!

packages/react-centra-checkout/src/PaymentForm/index.tsx
```diff
@@ -0,0 +1,84 @@
...
+  // POST to /payment if payment method is initiate only (method that doesn't need address) or if the payment method provides it's own address
+  React.useEffect(() => {
+    if (
+      selection &&
+      paymentMethod &&
+      (paymentMethod.providesCustomerAddressAfterPayment || paymentMethod.supportsInitiateOnly)
+    ) {
+      const { address, shippingAddress } = selection
+
+      submitPayment?.({
```
maeertin, line 36: How come this is needed as optional chaining @alexanderflink ?

janmysiak: image

alexanderflink: This is sort of typescript specific, but it is because none of the properties of the useCentra are actually guaranteed to exist. For example, if you try to access `submitPayment` but haven't wrapped your component in the `CentraProvider`. I think it's nice though, this ensures your app won't crash if you miss this.

maeertin: Awesome, thanks!

alexanderflink: image

PR #47, opened 2022-02-23
feature/wordpress-rest-api

maeertin:
# Todo's

- [ ] **Write tests for `ApiClient` based upon the existing one in `@oakwood/accelerator`.**
*Why?* you might ask. We want to have an equal interface as the old one, in regards for easy migration for all the old projects.

**Though**, the `@oakwood` version of the `ApiClient` may obviously have flaws and that has to be taken into account when completing this version.

- [ ] **Make sure they all tests pass**

### DIFF
--- a/packages/api-client/src/index.ts
+++ b/packages/api-client/src/index.ts
@@ -3,7 +3,7 @@ interface Options {
 }
 
 class ApiClient {
-  baseUrl?: string
+  baseUrl?: URL
 
   static default: ApiClient
 
@@ -22,7 +22,7 @@ class ApiClient {
   }
 
   constructor(baseUrl?: string, defaultRequestOptions?: RequestInit, settings?: Options) {
-    this.baseUrl = baseUrl
+    this.baseUrl = baseUrl ? new URL(baseUrl) : undefined
     this.defaultRequestOptions = { ...this.defaultRequestOptions, ...defaultRequestOptions }
     this.settings = { ...this.settings, ...settings }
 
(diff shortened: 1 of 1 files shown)


packages/api-client/package.json
```diff
@@ -0,0 +1,44 @@
+{
+  "name": "@noaignite/api-client",
```
maeertin, line 2: Is this supposed to be `@noaignite/wordpress-rest-api`? :)

packages/api-client/package.json
```diff
@@ -0,0 +1,44 @@
...
+    "api",
+    "client"
+  ],
+  "license": "MIT",
+  "main": "./src/index.ts",
+  "module": "./src/index.ts",
+  "repository": {
+    "type": "git",
+    "url": "https://github.com/noaignite/accelerator.git",
+    "directory": "packages/wordpress-rest-api"
```
maeertin, line 18: This should probably be `@noaignite/api-client` :)

alexanderflink:
@maeertin can you make this a draft? This one isn't ready to be merged yet I think.

maeertin:
Perhaps we could get assistance from @adamsoderstrom here? :)

adamsoderstrom:
> Perhaps we could get assistance from @adamsoderstrom here? :)

I would gladly help out, @maeertin!
Are there any specified goals for this PR? :)

alexanderflink:
Closed because Wordpress is no longer relevant for our new projects. We should look into refactoring the previous wordpress projects separately.

PR #49, opened 2022-03-04
fix(PaymentEmbed): add 'additionalPaymentProps'

janmysiak:
Not sure if this is the best solution but this field seems to be required when making a [/payment](https://docs.centra.com/swagger-ui/?api=CheckoutAPI#/4.%20selection%20handling%2C%20checkout%20flow/post_payment) request.


alexanderflink (commented):
I think this should probably be handled within the app, since I think it is a setting if `termsAndConditions` should be required or not.

PR #58, opened 2022-04-15
react-centra-checkout - Use cookie instead of localStorage for session token

alexanderflink:
This pull request changes the storage of the Centra checkout api token to a cookie instead of localStorage. This has a number of benefits:

- With a cookie, it's possible to read the token both server and client side. This means we can, for example, fetch the users selection directly on the server on the checkout and receipt pages, displaying a complete page instead of serving an empty page that is populated client side.
- Makes it possible to read the users session token in a Next.js middleware which opens up possibilities to do rewrites / redirects etc based on the token
- When searching for this topic, the general consensus seems to be storing session tokens in localStorage is not the way to go. Security is one of the reasons. While we will still have the same security concerns regarding XSS since this PR does not use a Http only cookie, this at least enables us to move in the right direction in that regard. If security at some point becomes an issue, we can move to using a Http only cookie and have some small proxy inbetween the Centra checkout api which reads the cookie and sends it as a header.

The only drawback is we introduce a small (780B) dependency on js-cookie to read cookies.

Along with the cookie functionality is a few small tweaks:
- `disableInitialSelection` renamed to `disableInit` since it doesn't actually disable the initial selection (which is an empty selection), but rather disables the init function from running
- added `initialSelection` which allows to pass an initial selection object to the Context. Useful if you for example fetch the selection server side

### DIFF
--- a/packages/react-centra-checkout/package.json
+++ b/packages/react-centra-checkout/package.json
@@ -46,5 +46,9 @@
   },
   "engines": {
     "node": ">=12.0.0"
+  },
+  "dependencies": {
+    "@types/js-cookie": "^3.0.1",
+    "js-cookie": "^3.0.1"
   }
 }
--- a/packages/react-centra-checkout/src/Context/index.tsx
+++ b/packages/react-centra-checkout/src/Context/index.tsx
@@ -1,16 +1,33 @@
 import * as React from 'react'
 import * as Centra from '@noaignite/centra-types'
+import cookies from 'js-cookie'
 import ApiClient from '../ApiClient'
 
 const apiClient = ApiClient.default
 
 /** The prop types that CentraProvider accepts */
 export interface ProviderProps {
+  /** Centra API URL */
   apiUrl: string
   children: React.ReactNode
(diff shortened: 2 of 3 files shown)


maeertin:
Awesome @alexanderflink ✨ Just one question, are cookie libraries still needed today 2022 for doing something as trivial as get, set & remove?

alexanderflink:
> Awesome @alexanderflink ✨ Just one question, are cookie libraries still needed today 2022 for doing something as trivial as get, set & remove?

Actually I just found there is a new API called [Cookie store](https://developer.mozilla.org/en-US/docs/Web/API/Cookie_Store_API) for dealing with cookies, but it's [not well supported](https://caniuse.com/cookie-store-api) at all. Bringing in a polyfill for this would also be a slightly larger dependency than js-cookie. Not sure whats best?

maeertin (approved):

maeertin:
Sounds good to me then @alexanderflink :) Approved ✅

janmysiak (approved):

PR #68, opened 2022-08-05
feat: add root level typescript support to monorepo

maeertin:
This PR adds the following:

Packages now have:
* Locally defined `tsconfig.json`, sets up development in typescript.
* Locally defined `tsconfig.build.json`, used for building type definitions.
* A `test` and a `typecheck` script.
* Updated `build` script which now transpiles to `modern`, `node` & `stable` (default) bundles. This new build script now uses the root `.browserslistrc` to set the babel target.

Migration of `packages/utils` to typescript.

New workspace root scripts added, `yarn test` & `yarn typecheck`.


alexanderflink:
@maeertin Looks very nice! Everything builds nicely on my machine as well. One question: how are consumers of the packages supposed to opt in / out of the "modern" or "stable" versions? Stable is set as `module` in package.json so that would be the default, but how would you opt in to the "modern" packages?

maeertin:
@alexanderflink Other bundles can always be opted in as following https://mui.com/material-ui/guides/minimizing-bundle-size/#available-bundles

PR #72, opened 2022-10-16
Add support for customising cookie options

Thomas0c:


packages/react-centra-checkout/src/Context/index.tsx
```diff
@@ -31,6 +31,12 @@ export interface ProviderProps {
     @defaultValue `centra-checkout-token`
   */
   tokenName?: string
+  tokenCookieOptions?: {
+    path?: string
+    domain?: string
+    secure?: boolean
+    sameSite?: 'Lax' | 'Strict' | 'None'
+  }
```
alexanderflink, line 39: Can we change this to be:
`tokenCookieOptions?: Cookies.CookieAttributes`

You can import the Cookie type like:
`import type Cookies from 'js-cookie'`

alexanderflink (changes requested):

packages/react-centra-checkout/src/Context/index.tsx
Thomas0c: Yes of course. Makes a lot of sense.

PR #79, opened 2022-11-04
Accept header token

Thomas0c:


adamsoderstrom:
This PR minimizes the occurrences of prefetches, since the request to Centra will only contain standard headers (`Accept`) and not custom one (`api-token`).

We've not yet seen that Centra mentions the `Accept` header to do this, but we'll have to try it out.

PR #80, opened 2022-11-04
Add support for passing tokenName to useCentraOrders

Thomas0c:
Experiencing that initial call to `useCentraOrders` returns 403 (access denied) given that the api-token header is not set unless one or more other calls have been made prior. Thus allowing for passing a token name and reading this, when calling `useCentraOrders`.

### DIFF
--- a/packages/react-centra-checkout/src/Context/index.tsx
+++ b/packages/react-centra-checkout/src/Context/index.tsx
@@ -637,10 +637,21 @@ export function useCentraReceipt(token: string): Centra.CheckoutApi.OrderComplet
   @param from - Display orders from this index. Defaults to 0.
   @param size - Display this many orders. Defaults lists all orders.
 */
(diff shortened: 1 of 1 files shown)


packages/react-centra-checkout/src/Context/index.tsx
```diff
@@ -636,21 +636,54 @@ export function useCentraReceipt(token: string): Centra.CheckoutApi.OrderComplet
...
 */
-export function useCentraOrders(from?: number, size?: number): Centra.CheckoutApi.OrdersResponse {
+export function useCentraOrders(
+  from?: number,
+  size?: number,
+  token?: {
+    tokenName?: string
+    tokenExpires?: number | Date
+    tokenCookieOptions?: Cookies.CookieAttributes
+  },
```
adamsoderstrom, line 650: **question:** Is this necessary?
**suggestion:** Would be nice if we can enforce that `tokenName` in both the `Context` as well as inside `useCentraOrders`.

PR #86, opened 2023-01-19
feat(ScrollProgress): add requestAnimationFrame and a new friction prop

maeertin:
Now with requestAnimationFrame for better performance and a new friction prop which adds easing.

BREAKING CHANGE: The bounds key has now been removed from the onChange callback arguments.

### DIFF
--- a/packages/oui/src/ScrollProgress/ScrollProgress.js
+++ b/packages/oui/src/ScrollProgress/ScrollProgress.js
@@ -12,21 +12,53 @@ export function calculateVerticalProgress(bounds, topOffset = 0, bottomOffset = 
 }
 
 const ScrollProgress = React.forwardRef(function ScrollProgress(props, ref) {
-  const { onChange, onEnter, onExit, ...other } = props
+  const { friction = 0, onChange, onEnter, onExit, precision = 0.001, ...other } = props
 
   const rootRef = React.useRef(null)
 
+  const stateRef = React.useRef({
+    currentInnerProgress: 0,
+    currentProgress: 0,
+    isEasing: false,
+    targetInnerProgress: 0,
+    targetProgress: 0,
+  })
+
+  const handleRaf = React.useCallback(() => {
+    const deltaIP = stateRef.current.targetInnerProgress - stateRef.current.currentInnerProgress
+    const deltaP = stateRef.current.targetProgress - stateRef.current.currentProgress
+
+    stateRef.current.currentInnerProgress += deltaIP * (1 - friction)
+    stateRef.current.currentProgress += deltaP * (1 - friction)
+
+    stateRef.current.isEasing = Math.abs(deltaIP) > precision || Math.abs(deltaP) > precision
+
+    if (stateRef.current.isEasing) {
+      requestAnimationFrame(handleRaf)
+    }
+
+    if (onChange) {
+      onChange({
+        innerProgress: stateRef.current.currentInnerProgress,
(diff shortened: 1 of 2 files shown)


packages/oui/src/ScrollProgress/ScrollProgress.js
```diff
@@ -12,21 +12,53 @@ export function calculateVerticalProgress(bounds, topOffset = 0, bottomOffset =
...
 
+  const stateRef = React.useRef({
+    currentInnerProgress: 0,
+    currentProgress: 0,
+    isEasing: false,
+    targetInnerProgress: 0,
+    targetProgress: 0,
+  })
+
+  const handleRaf = React.useCallback(() => {
```
adamsoderstrom, line 27: I would call this function one of the following: 

```suggestion
  const handleAnimationFrame = React.useCallback(() => {
```
```suggestion
  const handleAf = React.useCallback(() => {
```

I think `Raf` stands for `requestAnimationFrame`, but i think the `requestAnimationFrame` is a function requesting the actual animation frame.

I think that the name `handleRaf` sounds like a wrapper for the actual `requestAnimationFrame` function and no a callback for it.

packages/oui/src/ScrollProgress/ScrollProgress.js
```diff
@@ -12,21 +12,53 @@ export function calculateVerticalProgress(bounds, topOffset = 0, bottomOffset =
...
+  const { friction = 0, onChange, onEnter, onExit, precision = 0.001, ...other } = props
 
   const rootRef = React.useRef(null)
 
+  const stateRef = React.useRef({
+    currentInnerProgress: 0,
+    currentProgress: 0,
+    isEasing: false,
+    targetInnerProgress: 0,
+    targetProgress: 0,
```
adamsoderstrom, line 24: I don't really understand what `current` and `target` represents.

`target` seems to be the actual scroll progress, while `current` seems to be a modified version `target` (because of the `friction` property)..?

I wonder if we could make the distinction even clearer here.

packages/oui/src/ScrollProgress/ScrollProgress.js
maeertin: Sure thing, could go for `handleAnimationFrame` :)

packages/oui/src/ScrollProgress/ScrollProgress.js
adamsoderstrom: Yeah, as we discussed in person today; it's a bit hard to distinguish what `current` and `target` actually refers to.

Let's let it marinate a while! 😊

maeertin: My suggestion would be to go with the following:

```js
const stateRef = React.useRef({
    isEasing: false,
    rafInnerProgress: 0,
    rafProgress: 0,
    scrollInnerProgress: 0,
    scrollProgress: 0,
})

const handleRaf = () => {
  // updates raf values
}

const handleScroll = () => {
  // updates scroll values
}
```

With this change, the state names match up with the naming of the handlers, what do you think @adamsoderstrom? :)

adamsoderstrom: Ah, nice! That looks like a nice tradeoff! 🎉

adamsoderstrom (approved):

alexanderflink (approved):

PR #92, opened 2023-02-20
Feature/centra types update

alexanderflink:
- Rewrite centra-types to es modules instead of global type definitions generated from swagger
- Update react-centra-checkout to the new types


janmysiak (approved):
Looks good to me!

If anything we might want to type nested objects/arrays separately, as this may lead to some convoluted code. It's not uncommon for us to work with [items](https://github.com/noaignite/accelerator/pull/92/files#diff-79c86415da37ee2931a31873adddb5f2bf887b65b0f5d4a8eb8eed5063897010R69) in isolation (for example).

Although I feel this can be done in another PR.

adamsoderstrom (approved):
LGTM!

In regards to doing this during client-specific projects, i think we'll have to add types here "as we go".

packages/centra-types/src/models/index.ts
```diff
@@ -1,53 +1,53 @@
...
 
-export { default as Location } from './Location'
+export type { default as Location } from './Location'
 
-export { default as LoggedIn } from './LoggedIn'
+export type { default as LoggedIn } from './LoggedIn'
 
-export { default as Order } from './Order'
+export type { default as Order } from './Order'
 export * from './Order'
```
adamsoderstrom, line 35: Lovely!

adamsoderstrom (approved):

PR #93, opened 2023-03-02
Feature/api client prop

alexanderflink:
This feature adds the possibility to pass a different ApiClient than the default one to `CentraProvider`. This makes things like doing custom requests, formatting data and a lot of other things possible.

See [cake](https://github.com/noaignite/cake/blob/4d566d638117d2254becf5932b6a900e3e06c882/src/contexts/Root/index.js#L47) for an example of how this works. Cake uses TRPC, but the ApiClient could do anything really.

### DIFF
--- a/packages/react-centra-checkout/src/Context/index.tsx
+++ b/packages/react-centra-checkout/src/Context/index.tsx
@@ -11,9 +11,9 @@ const centraEvents = CentraEvents.default
 /** The prop types that CentraProvider accepts */
 export interface ProviderProps {
   /** Centra API URL */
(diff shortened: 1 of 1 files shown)


axeen (commented):
I assume all the callback deps for `apiClient` are necessary?

alexanderflink:
> I assume all the callback deps for `apiClient` are necessary?

Yes unfortunately, It's because if the apiClient prop is changed, all of the callbacks need to re-render.

adamsoderstrom (approved):
LGTM, @alexanderflink !

PR #97, opened 2023-03-24
docs(Formit): add 'Sourcing values from API call' example

adamsoderstrom:
Added a section / example of how to source (somewhat) initial values from an API.

My initial alternative here may not be best practice. If not, then i'm creating this PR to spark the discussion on how the example should be.

### DIFF
--- a/packages/formit/README.md
+++ b/packages/formit/README.md
@@ -80,6 +80,62 @@ const BasicExample = () => {
 export default BasicExample
 ```
 
+#### Sourcing values from API call
+
+Sometimes, the values you want to edit via Formit originates from an API.
+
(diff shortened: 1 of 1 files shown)


maeertin (changes requested):
Awesome addition to the README @adamsoderstrom ✨ You can achieve this behaviour of reinitializing the `initialValues` in formit by passing `enableReinitialize: true`. This will make it so that formit reinits when the `initialValues` object changes. Do you want to try and rewrite this section of the README? 😄

adamsoderstrom:
Ah, thanks, @maeertin!
I would gladly try to look at this later on! 😊

maeertin (approved):
@adamsoderstrom you're on 🔥 Thank you for the contribution!

PR #102, opened 2023-04-11
feature/color-contrast

Dashice:
Adds a `colorContrast` function, which accepts a HEX `baseColor`, proceeded with any number of HEX `colors` and returns an object containing:

```tsx
contrast: string; // The most contrasting color in the `color` arguments passed
contrastRatio: number; // The contrast ratio for the `contrast` color (the most contrasting color in relation to the `baseColor`)
fallbackColor: string; // Either #000 or #fff, the most appropriate fallback color to use if the `contrastRatio` is not sufficient
```

### DIFF
--- a/packages/utils/src/colorContrast.test.js
+++ b/packages/utils/src/colorContrast.test.js
@@ -1,0 +1,39 @@
+import colorContrast from './colorContrast'
+
+describe('colorContrast', () => {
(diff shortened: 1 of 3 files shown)


janmysiak (approved):
Looks good, well done!

maeertin (approved):
Looking good, let's ship it!! 🛥️

PR #109, opened 2024-01-16
add `engine-strict` to `.npmrc`

adamsoderstrom:
# Background

Prior to this PR, you could experience errors like this when running `pnpm build`:

```bash
web:build:
web:build: > web@1.0.2 build /Users/adamsoderstrom/Documents/NoA/Repos/accelerator/apps/web
web:build: > next build
web:build:
web:build: unhandledRejection ReferenceError: Headers is not defined
web:build:     at Object.<anonymous> (/Users/adamsoderstrom/Documents/NoA/Repos/accelerator/node_modules/.pnpm/next@14.0.3_react-dom@18.2.0_react@18.2.0/node_modules/next/dist/server/web/spec-extension/adapters/headers.js:32:30)
...
```

It doesn't make much sense. 
One could start digging into what module is missing and why.

On the other hand, one could look in the `package.json` and see the supported Node version.

```json
  "engines": {
    "node": ">=18"
  }
```

Though, time is often already wasted when reaching for answers in the Node version.

# Results

Now, you can't even execute `pnpm install` if you're using an unsupported Node version. See the following example output:

```bash
pnpm install
 ERR_PNPM_UNSUPPORTED_ENGINE  Unsupported environment (bad pnpm and/or Node.js version)

Your Node version is incompatible with "/Users/adamsoderstrom/Documents/NoA/Repos/accelerator".

Expected version: >=18
Got: v16.15.1

This is happening because the package's manifest has an engines.node field specified.
To fix this issue, install the required Node version.
```

# References

- [`engine-strict`](https://docs.npmjs.com/cli/v9/using-npm/config#engine-strict) flag


maeertin (approved):
LGTM ✨

PR #111, opened 2024-02-11
add stylex to `ui` package

adamsoderstrom:
Needs to explore and ensure wanted features are working as expected (CSS variables, etc.).


adamsoderstrom:
Stylex's documentation recommend a Babel plugin for [development server](https://stylexjs.com/docs/learn/installation/#development) but unsure about if that's a direction one wants to take, since "dev mode" injects the styling in the JavaScript modules and not to a CSS file.

In production mode, one would have to import a CSS file from the `ui` package. While in development mode, one _can't_ rely on the CSS file, since dev mode removes it and relies on the inline JS styles.

The current solution in this PR _works_, but may want improvements in terms of DX.
If has both the `web` app and the `ui` package in dev mode, then the `web` app will throw errors upon file changes to the resulting CSS file.

adamsoderstrom:
Closing, since [StyleX forces consumers to also use StyleX](https://github.com/facebook/stylex/discussions/442).

PR #118, opened 2024-04-04
feat(react-centra-checkout): add `removeVouchers` handler

adamsoderstrom:


adamsoderstrom:
What do you say? We could modify the already existing `removeVoucher` handler to support lack of argument (and then remove all vouchers, since the API behaves that way).

janmysiak:
LGTM!

Think I prefer this solution over updating the existing handler. It would feel unintuitive that 'removeVoucher' would be used for removing **all** vouchers from the selection (its name implies it removes a single voucher).

Minor nitpick, but maybe it'd be worth renaming the added handler to 'removeAllVouchers' or something. Currently, it sounds to me like it would be used to remove multiple vouchers.

PR #119, opened 2024-04-09
Next upgrade deps

maeertin:


package.json
```diff
@@ -4,21 +4,30 @@
   "scripts": {
     "build": "turbo build",
     "dev": "turbo dev",
+    "prettier": "prettier --check .",
+    "prettier:fix": "prettier --write .",
     "lint": "turbo lint",
-    "test": "vitest run",
-    "format": "prettier --write \"**/*.{ts,tsx,md}\"",
+    "test": "pnpm run lint && pnpm run type-check",
```
adamsoderstrom, line 10: **fix:** run `vitest` on `pnpm run test`

There's an example test in the `ui` package. Looks like that won't execute.

adamsoderstrom (changes requested):

package.json
maeertin: The `vitest` test in `ui` was just a "lorem ipsum" test that we can remove for now :)

maeertin: This has now been removed 👍

Dashice (approved):

PR #121, opened 2024-04-10
feat: add new `@noaignite/tailwind-typography` package

maeertin:


packages/tailwind-typography/src/index.ts
```diff
@@ -0,0 +1,205 @@
...
+export type TypographyOptionsVariantMap<TKeys extends readonly [string, ...string[]]> = {
+  clamp?: boolean
+  clampMax?: boolean
+  clampMin?: boolean
+  fluid?: boolean
+} & {
+  [key in TKeys[number]]?: TypographyOptionsVariant
+} & {
+  [key in TKeys[0]]-?: TypographyOptionsVariant
+}
```
adamsoderstrom, line 69: **thought:** the intersection between different keys seem a bit confusing

The following intersection:

```typescript
{
  [key in TKeys[number]]?: TypographyOptionsVariant
} & {
  [key in TKeys[0]]-?: TypographyOptionsVariant
}
```

Seems a bit hard to follow. Maybe you could add a comment to describe that the first entry in`TKeys` will result in a "required" `TypographyOptionsVariant`?

packages/eslint-config/library.js
adamsoderstrom, line 1: **todo:** move ESlint config changes to separate PR

I understand the PR is dependent on this config change.
Though, would be nice to evaluate the new package in this PR, while evaluate results of the ESlint config change on the already added packages in isolation.

packages/tailwind-typography/package.json
```diff
@@ -0,0 +1,48 @@
...
+    "url": "https://github.com/noaignite/accelerator.git",
+    "directory": "packages/tailwind-typography"
+  },
+  "bugs": {
+    "url": "https://github.com/noaignite/accelerator/issues"
+  },
+  "scripts": {
+    "build": "tsup",
+    "dev": "tsup --watch",
+    "lint": "eslint . --max-warnings 0",
```
adamsoderstrom, line 28: **praise:** Nice to see the `--max-warnings` flag after our discussion!

packages/tailwind-typography/src/index.ts
```diff
@@ -0,0 +1,205 @@
+import plugin from 'tailwindcss/plugin'
+
+function isPlainObject(x: unknown): x is Record<string, unknown> {
+  return (
+    typeof x === 'object' &&
+    x?.constructor === Object &&
+    Object.getPrototypeOf(x) === Object.prototype
+  )
+}
```
adamsoderstrom, line 9: **thought:** we should make an internal package of "type predicate" functions

Like this `isPlanObject` function.

packages/tailwind-typography/src/index.ts
```diff
@@ -0,0 +1,205 @@
...
+  fontWeight: number,
+  fontSize: number,
+  lineHeight: number,
+  letterSpacing: number,
+  other?: Record<string, unknown>,
+) => ({
+  fontFamily,
+  fontWeight,
+  fontSize,
+  // Unit-less following http://meyerweb.com/eric/thoughts/2006/02/08/unitless-line-heights/
```
adamsoderstrom, line 22: **todo:** move this `line-height` comment to the type declaration as JsDoc

So consumers can get more context.

packages/tailwind-typography/src/index.ts
```diff
@@ -0,0 +1,205 @@
...
+          const variant = variantOrMap[breakpoint]
+          const width: string = theme(`screens.${breakpoint}`)
+          const media = `@media (min-width: ${width})`
+
+          if (!fluid) {
+            const fontSize = `${variant.fontSize}${unit}`
+            let variantStyles = acc[variantName]
+
+            if (idx === 0) {
+              variantStyles = { ...variant, fontSize }
```
adamsoderstrom, line 154: **thought:** does this work?

I imagine that this will result in the following:

```
> const acc = {}
> let variantName = 'xs'
> let variantStyles = acc[variantName] // `undefined`
> variantStyles = {bar: true}
> acc[variantName] === variantStyles // `false`
```

Since `undefined` doesn't result in a reference to `acc`.

adamsoderstrom (commented):
Great work, @maeertin!

packages/tailwind-typography/src/index.ts
maeertin: Nice catch! Now fixed :)

packages/tailwind-typography/src/index.ts
maeertin: Good point! Although I'll be simplifying this even further and remove the `buildTypographyVariant' as it's really not needed.

packages/eslint-config/library.js
maeertin: Agreed 👍

packages/tailwind-typography/src/index.ts
maeertin: Totally agree, comments coming up :)

PR #123, opened 2024-04-11
chore: add `@noaignite/utils` to Turborepo setup

adamsoderstrom:
In this PR, we're migrating the `@noaignite/utils` package from the old monorepo setup to use the Turborepo setup that's currently established in the `next` branch.

There are some source code modifications in this PR compared to what's currently in the `main` branch:
- [Casting to `as [number, number, number]` on result of `.map` call](https://github.com/noaignite/accelerator/pull/123/files#diff-e7917f240151fbb5b5793155fecf76e954042b94cc8d5ccb2568bb37b832ccfeR8) - This is due to the [`noUncheckedIndexedAccess`](https://www.typescriptlang.org/tsconfig#noUncheckedIndexedAccess) rule inherited from the updated TS config.
- [The replacement of the classic for loop to a `for of` loop](https://github.com/noaignite/accelerator/pull/123/files#diff-e7917f240151fbb5b5793155fecf76e954042b94cc8d5ccb2568bb37b832ccfeR66) - This is due to the [`noUncheckedIndexedAccess`](https://www.typescriptlang.org/tsconfig#noUncheckedIndexedAccess) rule inherited from the updated TS config.

Besides those changes; if human error has occurred, then that could become more obvious when merging the `next` branch into `main` later on.


packages/utils/package.json
```diff
@@ -0,0 +1,41 @@
...
+    "type": "git",
+    "url": "https://github.com/noaignite/accelerator.git",
+    "directory": "packages/utils"
+  },
+  "bugs": {
+    "url": "https://github.com/noaignite/accelerator/issues"
+  },
+  "scripts": {
+    "build": "tsup",
+    "test": "cd ../../ && cross-env NODE_ENV=test jest packages/utils",
```
adamsoderstrom, line 25: **fix:** Change to `vitest`

packages/utils/README.md
```diff
@@ -0,0 +1,15 @@
...
+
+Utils is available as an [npm package](https://www.npmjs.com/package/@noaignite/utils).
+
+```sh
+// with npm
+npm install @noaignite/utils
+
+// with yarn
+yarn add @noaignite/utils
+```
```
Dashice, line 15: Should we omit mentions of `npm` in favor of `pnpm` or do we just add `pnpm` here.
@maeertin @adamsoderstrom

Dashice (approved):

PR #125, opened 2024-04-11
chore: add `@noaignite/create-app` to Turborepo setup

adamsoderstrom:
In this PR, we're migrating the `@noaignite/create-app` package from the old monorepo setup to use the Turborepo setup that's currently established in the `next` branch.

There shouldn't be any source code modifications in this PR compared to what's currently in the `main` branch. If human error has occurred, then that could become more obvious when later on merging `next` branch into `main`.


packages/create-app/package.json
```diff
@@ -0,0 +1,46 @@
...
+  },
+  "bugs": {
+    "url": "https://github.com/noaignite/accelerator/issues"
+  },
+  "scripts": {
+    "build": "pnpm prebuild && mkdir dist && cp -a package.json LICENSE CHANGELOG.md README.md createApp.js index.js dist/",
+    "prebuild": "rm -rf dist",
+    "release": "pnpm build && npm publish ./dist --tag latest",
+    "release:yalc": "pnpm build && yalc publish ./dist"
+  },
```
adamsoderstrom, line 31: **thought:** Remove these

Since we don't use the `dist` directory at all. We simply target the file to execute by applying `./index.js` to the `bin.createApp` property.

PR #127, opened 2024-04-11
docs(utils): add `pnpm` example

adamsoderstrom:


maeertin:
Should this PR maybe add a `pnpm install @noaignite/*` example for all packages?

PR #128, opened 2024-04-12
fix(ci): `type-check` after `build`

adamsoderstrom:
The CI pipeline currently breaks while running `pnpm run ci` with the following error: 

```bash
Error: app/page.tsx(1,36): error TS2307: Cannot find module '@repo/ui' or its corresponding type declarations.
```

This is caused by the execution order in the `ci` script:

```
pnpm run lint && pnpm run test && pnpm run build
```

We can't run `test` before `build`.
`test` runs `tsc` which requires that all types are accessible in the current package, which never does due to that `pnpm run build` hasn't yet executed.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -14,7 +14,7 @@ jobs:
       - uses: actions/checkout@v3
       - uses: pnpm/action-setup@v2
         with:
-          version: 8
+          version: 9
       - uses: actions/setup-node@v3
         with:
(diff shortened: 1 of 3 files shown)


adamsoderstrom:
@maeertin: change to `pnpm run build && pnpm run test` in `ci` script.

maeertin (approved):
Nice one @adamsoderstrom 🎉

PR #133, opened 2024-08-01
New monorepo setup made with turborepo

maeertin:


maeertin:
The migration PRs for the old packages have now been finalized, these being:
* `@noaignite/centra-types`
* `@noaignite/react-centra-checkout`
* `@noaignite/utils`

I've also tried publishing these built bundles to a local `verdaccio` registry and then installed them on one of our projects which seems to work as intended.

Question: Is this a major/minor/patch type of update? They are currently marked as major updates by changeset as that was the default behaviour as we migrated to `changeset` from `conventional-commits`. This can however be changed if we see fit.

alexanderflink (approved):

PR #142, opened 2024-08-08
chore(react-centra-checkout): change centra-types to workspace dependency

alexanderflink:
I think this solves the issue where react-centra-checkout is trying to install a new version of `centra-types` from npm which doesn't exist yet because it hasn't been published. I got the same error locally when running `pnpm install` as in CI workflow, but this fixed it for me locally.


maeertin:
Closing as the solution was to add `link-workspace-packages=true` to `.npmrc`.

PR #150, opened 2024-08-12
feat(react-centra-checkout): return promise for `resetSelection`

### LINKED ISSUE #107: [react-centra-checkout]: `resetSelection` race condition
It appears that the `resetSelection` centra handler isn't asynchronous, when i believe it'd benefit from being so.
The `resetSelection` removes the `centra-checkout-token` cookie and calls the `init` handler. The `init` call isn't awaited, resulting in a new token being used inside the app-global `CentraContext` upon resolution of the function call. Though, if one would perform page route change (via `router.push` or similar), you'd potentially apply a different cookie than that's used inside the `CentraContext`. For example, when you're utilizing the Next.js middleware of redirects or middleware parameters.

adamsoderstrom:
fixes #107


adamsoderstrom:
Closing, in favor of #153

PR #153, opened 2024-08-12
fix/centra checkout typescript eslint disable comments

adamsoderstrom:
---
- **chore(updateCartItemSize): resolve `eslint-disable-next-line` comments**
- **chore(centraCheckoutCallback): resolve `eslint-disable-next-line`**
- **fix(react-centra-checkout): return void from `init` onMount of `Context`**
- **feat(react-centra-checkout): extend `GlobalEventHandlersEventMap`**
- **fix(centraCheckoutCallback): resolve `@typescript-eslint` disable comments**
- **fix(useCentraReceipt): resolve `@typescript-eslint` disable comments**
- **chore(useCentraOrders): resolve `@typescript-eslint` disable comments**


alexanderflink (approved):
LGTM

packages/react-centra-checkout/src/types/global.d.ts
```diff
@@ -6,3 +6,7 @@ interface Window {
     reInitiate: (plugin: string) => void
   }
 }
+
+interface GlobalEventHandlersEventMap {
+  centra_checkout_callback: CustomEvent<Record<string, unknown> | undefined>
+}
```
maeertin, line 12: question: The only thing I'm wondering if this would need to be exported for when a consumer of this package subscribes to our custom events. Do you perhaps know?

adamsoderstrom: It could.
Though, i wonder if it would be easier to apply the types on a per-project basis, if one want to manually listen for `centra_checkout_callback` events.

adamsoderstrom:
Tested locally in client project and generic customer flow was done successfully.

packages/react-centra-checkout/src/PaymentEmbed/index.tsx
```diff
@@ -10,6 +10,23 @@ export interface PaymentEmbedProps {
...
+  for (const key in value) {
+    // Check if the key is a string and the value is a string
+    if (typeof key !== 'string' || typeof value[key as keyof typeof value] !== 'string') {
+      return false
+    }
+  }
+
+  return true
+}
+
```
adamsoderstrom, line 29: Maybe this would be worth putting in another package? @maeertin @alexanderflink

packages/react-centra-checkout/src/PaymentEmbed/index.tsx
```diff
@@ -75,8 +90,11 @@ const PaymentEmbed = memo(function PaymentEmbed(
         })
         .catch((err: unknown) => {
           console.error('@noaignite/react-centra-checkout: Could not submit payment')
-          // @ts-expect-error -- TODO: Fix this
-          setPaymentCallbackError(err)
+
+          if (isRecordStringString(err)) {
```
adamsoderstrom, line 94: The reason were using `isRecordStringString` function, instead of a `isCheckoutApiError` is to avoid false positives if the `CheckoutApi.Error` type would be changed in the future. 
The `CheckoutApi.Error` is equal to a `Record<string,string>` today.

packages/react-centra-checkout/src/ApiClient/index.ts
```diff
@@ -26,10 +26,8 @@ class ApiClient {
       body: ['POST', 'PUT'].includes(method) ? JSON.stringify(data) : undefined,
     })
 
-    // eslint-disable-next-line @typescript-eslint/no-unsafe-assignment -- TODO: fix this
-    const json = await response.json()
+    const json = (await response.json()) as unknown
```
adamsoderstrom, line 29: This will result in that callees of `ApiClient.request` will have to parse / validate / cast the returned value.
Doing this will result in more explicitness and enforcing stronger type safety.

packages/react-centra-checkout/src/Context/index.tsx
```diff
@@ -232,34 +232,35 @@ export function CentraProvider(props: ProviderProps) {
...
-      if ('selection' in response && response.selection) {
-        setSelection(response)
+      if (
+        Boolean(response) &&
+        response !== null &&
+        typeof response === 'object' &&
+        'selection' in response &&
+        response.selection
+      ) {
+        setSelection(response as CheckoutApi.SuccessResponse<CheckoutApi.SelectionResponse>)
```
adamsoderstrom, line 246: In result to `ApiClient.request` returning `Promise<unknown>`, we had to perform some additional type narrowing. This wasn't enforced by TS, since the the returned value was `Promise<any>`

packages/react-centra-checkout/src/PaymentEmbed/index.tsx
maeertin: question: Is a `isRecordStringString` check not too strict? What if the returned interface would look something like:
```ts
{
  status: 123,
  message: 'Lorem',
  other: 'Ipsum',
}
```
`status` being a number would fail the predicate. Do we not just want to check for `isPlainObject` and perhaps that a specific key exists or something similar?

packages/react-centra-checkout/src/PaymentEmbed/index.tsx
maeertin: Look at my comment [here](https://github.com/noaignite/accelerator/pull/153#discussion_r1714853602) :)

packages/react-centra-checkout/src/ApiClient/index.ts
maeertin: suggestion: I wonder if we should not just include Matt's `ts-reset` to get some fixes like this.

packages/react-centra-checkout/src/Context/index.tsx
maeertin: question: I understand the "if statement" refactor based off of the new type given to the `apiCall` argument. However, I'm not sure I follow as to why the type for `apiCall` was refactored to use `unknown` instead of `CheckoutApi.Response<CheckoutApi.SelectionResponse>`. Could you elaborate?

packages/react-centra-checkout/src/ApiClient/index.ts
adamsoderstrom: True, will do it in a separate PR, @maeertin!

packages/react-centra-checkout/src/Context/index.tsx
adamsoderstrom: The main reason is that we otherwise would have to do the "if statement" in all places where we call `selectionApiCall`.
We could though, move the "if statement" to a separate helper function, wrap the parameter passed to `selectionApiCall` in every call. By doing so, we could revert the changes made in `selectionApiCall`, @maeertin

packages/react-centra-checkout/src/PaymentEmbed/index.tsx
adamsoderstrom: I've now made changes to this and now we're using `isPlainObject`.

packages/react-centra-checkout/src/Context/index.tsx
alexanderflink: Chiming in here! I think we should do the if statement in a helper function like you said @adamsoderstrom! Because the `selectionApiCall` should actually expect an apiCall that returns a SelectionResponse, otherwise you could pass whatever into it, and then it doesn't make sense to call it a `selectionApiCall`

packages/react-centra-checkout/src/Context/index.tsx
```diff
@@ -232,34 +227,35 @@ export function CentraProvider(props: ProviderProps) {
...
+    async (apiCall: Promise<unknown> | (() => Promise<unknown>)) => {
       window.CentraCheckout?.suspend()
       const response = typeof apiCall === 'function' ? await apiCall() : await apiCall
 
-      if ('selection' in response && response.selection) {
-        setSelection(response)
+      if (
+        Boolean(response) &&
+        response !== null &&
+        typeof response === 'object' &&
```
maeertin, line 237: suggestion: As we now have a `isPlainObject` predicate in `@noaignite/utils` which already is a dependency of this package, we might as well replace these 3 lines with the predicate :)

packages/react-centra-checkout/src/ApiClient/index.ts
```diff
@@ -26,18 +26,22 @@ class ApiClient {
...
+    if (!ApiClient._default) {
+      ApiClient._default = new ApiClient()
+    }
+
+    return ApiClient._default
+  }
+
+  public get default(): ApiClient {
+    return ApiClient.default
+  }
```
maeertin, line 44: Nice one!

packages/react-centra-checkout/package.json
```diff
@@ -34,10 +34,12 @@
     "test-types": "tsc --noEmit"
   },
   "dependencies": {
+    "@noaignite/utils": "workspace:*",
```
maeertin, line 37: todo: Now that this is actually a dependency and not a devDep I wonder if this should not actually have the current version of the `@noaignite/types` package. The changeset-bot does actually update these via the `changeset version` script when it creates the `Version Packages` PR.

packages/react-centra-checkout/src/types/global.d.ts
```diff
@@ -1,8 +1,51 @@
...
+    // the interface that the centra checkout scripts adds to window
+    CentraCheckout?: {
+      suspend: () => void
+      resume: () => void
+      reInitiate: (plugin: string) => void
+    }
+  }
+
+  interface GlobalEventHandlersEventMap {
+    centra_checkout_callback: CustomEvent<Record<string, unknown> | undefined>
```
adamsoderstrom, line 44: **todo:** add `CentraCheckoutEventDetails` type here (available in `main`)

packages/react-centra-checkout/src/ApiClient/index.ts
adamsoderstrom, line 44: **todo:** Add documentation describing why there's two getters for this.

packages/react-centra-checkout/src/Context/index.tsx
adamsoderstrom: I've now done what you've mentioned above, @alexanderflink.
Though, i wonder how we want this design change to be exposed from the handlers as well?
For example, returning `Promise<unknown>` instead of `Promise<CheckoutApi.SuccessResponse<CheckoutApi.Selection>>`?

packages/react-centra-checkout/src/Context/index.tsx
adamsoderstrom: Done, @maeertin!

maeertin (approved):
LGTM

packages/react-centra-checkout/src/Context/index.tsx
adamsoderstrom: Consider my earlier comment out of scope for this PR.
This PR will not obfuscate the exposed return types of the handlers in `Context`.

PR #159, opened 2024-08-13
add ts reset

adamsoderstrom:
- **build(react-centra-checkout): add `@total-typescript/reset`**
- **refactor(ApiClient): remove `as unknown` casting**


alexanderflink:
Coud this possibly affect the types in any project that uses `react-centra-checkout`?

maeertin:
> Coud this possibly affect the types in any project that uses `react-centra-checkout`?

Hmm valid point, I actually don't know 🤔

packages/react-centra-checkout/src/ApiClient/index.ts
```diff
@@ -5,7 +5,7 @@ class ApiClient {
 
   options: RequestInit
 
-  static default: ApiClient
+  private static _default?: ApiClient
```
alexanderflink, line 8: Are we sure this should be private? What if you want to modify the default client instead of bringing your own?

adamsoderstrom: We'll use the getters to ensure both instance and class access will lead to the same instance of `ApiClient.default`.

adamsoderstrom:
> Coud this possibly affect the types in any project that uses `react-centra-checkout`?

True! Will have to look into it.

adamsoderstrom:
**todo:** `git rebase --onto` on target branch, since that branch has been rebased.

PR #169, opened 2024-08-14
build(react-utils): drop commonJS support

maeertin:
# Background
To my surprise I noticed that when consuming this package in a NextJS project the `next build` script fails for the `createRenderBlock` function meanwhile `next dev` works just fine. I'm still not entirely sure what the issue is, but I can get `next build` to succeed if I skip the `commonJS` build and only build `esm` modules.

# Changes
I discussed this with @adamsoderstrom and went ahead and dropped the support for `commonJS` for this package as this package is after all meant to be consumed in an `esm` environment. Furthermore, stable `esm` support has been in `node` since version 14 and latest node version is currently at 22. Additionally it's easier to debug the built files as the `dist` folder is less bloated.

### DIFF
--- a/.changeset/little-panthers-invite.md
+++ b/.changeset/little-panthers-invite.md
@@ -1,0 +1,5 @@
+---
+'@noaignite/react-utils': minor
+---
+
+drop support for commonJS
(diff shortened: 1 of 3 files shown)


adamsoderstrom (approved):
LGTM

PR #170, opened 2024-08-14
hotfix/remove-centra-checkout-callback-suspend

alexanderflink:
This fixes an issue we are having on a client project where `CentraCheckout.suspend()` is called after clicking the submit payment button in Klarna. This is apparently not supposed to be called according to [these](https://centra.dev/docs/tools/scripts/centra-checkoutscript#suspend-and-resume) docs.

Also fixes an unintentional bug in PaymentEmbed causing it not to render.

### DIFF
--- a/.changeset/seven-socks-help.md
+++ b/.changeset/seven-socks-help.md
@@ -1,0 +1,5 @@
+---
+'@noaignite/react-centra-checkout': patch
+---
+
+Removes invalid call to CentraCheckout.suspend, fixes PaymentEmbed not rendering
--- a/packages/react-centra-checkout/src/Context/index.tsx
+++ b/packages/react-centra-checkout/src/Context/index.tsx
@@ -259,6 +259,10 @@ export function CentraProvider(props: ProviderProps) {
         event.detail,
       )) as CheckoutApi.Response<CheckoutApi.SelectionResponse>
 
+      if ('selection' in response && response.selection) {
+        setSelection(response)
+      }
+
       window.CentraCheckout?.resume(event.detail.additionalFields?.suspendIgnore)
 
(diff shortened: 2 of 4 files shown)


adamsoderstrom (commented):
As i understand it, under "[Events](https://centra.dev/docs/tools/scripts/centra-checkoutscript#events)" we have an example of handling `centra_checkout_callback`.
According to the docs, When handling `centra_checkout_callback`, we shall **not** call `suspend` and **must** call `resume` after the request has resolved.

Under "[Suspend() and Ignore()](https://centra.dev/docs/tools/scripts/centra-checkoutscript#suspend-and-resume)", we see an example of handling quantity change in a cart, were we **must** call `suspend` before the request and we **must** call `resume` after the request has resolved.

If i've understood it correctly, then i think everything looks good here.

alexanderflink:
> As i understand it, under "[Events](https://centra.dev/docs/tools/scripts/centra-checkoutscript#events)" we have an example of handling `centra_checkout_callback`. According to the docs, When handling `centra_checkout_callback`, we shall **not** call `suspend` and **must** call `resume` after the request has resolved.
> 
> Under "[Suspend() and Ignore()](https://centra.dev/docs/tools/scripts/centra-checkoutscript#suspend-and-resume)", we see an example of handling quantity change in a cart, were we **must** call `suspend` before the request and we **must** call `resume` after the request has resolved.
> 
> If i've understood it correctly, then i think everything looks good here.

Yes, correct. This is also how I understand it, and have changed the handler accordingly.

adamsoderstrom:
> Yes, correct. This is also how I understand it, and have changed the handler accordingly.

Alright, will approve!

adamsoderstrom (approved):

PR #172, opened 2024-08-16
feat(types): add new Split type helper

maeertin:


adamsoderstrom (commented):
LGTM!

Though, i'm wondering if [`type-fest`](https://github.com/sindresorhus/type-fest)'s solution would be sufficient?

PR #173, opened 2024-08-16
feat(react-utils): add new createSvgIcon helper

maeertin:
There are many ways of working with svg's and so this helper is not intended to be the one and only solution. However, it is a handy little helper to quickly get started.

### DIFF
--- a/packages/react-utils/src/createSvgIcon.tsx
+++ b/packages/react-utils/src/createSvgIcon.tsx
@@ -1,0 +1,58 @@
+import {
+  forwardRef,
+  memo,
+  type ComponentPropsWithoutRef,
+  type ForwardedRef,
+  type ReactNode,
+} from 'react'
+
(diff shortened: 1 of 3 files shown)


packages/react-utils/src/createSvgIcon.tsx
```diff
@@ -0,0 +1,58 @@
...
+  type ForwardedRef,
+  type ReactNode,
+} from 'react'
+
+/**
+ * Create a new SVG icon component.
+ *
+ * @param path - The SVG path to render.
+ * @param displayName - The display name of the icon.
+ * @param viewBox - The viewBox of the SVG.
```
adamsoderstrom, line 14: **nitpick:** change to `viewBox`

adamsoderstrom (approved):
LGTM!
Have added one nitpick.

PR #174, opened 2024-08-19
feat(react-hooks): add new `react-hooks` package

maeertin:
Created an empty scaffold for a new `react-hooks` package where we can start gathering useful hooks we want to publish.


maeertin:
Closing this PR as we've decided make hooks part of the `react-utils` package.

PR #175, opened 2024-08-19
react-centra-checkout: Add ts reset

adamsoderstrom:


adamsoderstrom:
**todo:** Ensure that `ApiClient.request` returns `Promise<unknown>` from a consumer standpoint.

maeertin:
I've now verified that when using `ts-reset` within a package that these "fixed" typings are not exposed to the consumer of the package.

Package with `ts-reset`:
```ts
async function fetchSomething() {
  const response = await fetch('https://api.github.com/users/johndoe');
  const data = await response.json();

  return data; // typed as `unknown`
}
```

Consumer without `ts-reset`:
```ts
import { fetchSomething } from 'my-package'

async function Page() {
  const data = await fetchSomething();

  return data; // typed as `any`
}
```

maeertin (approved):
LGTM ✨

Issue #77, opened 2022-10-27
Create package for centra mock data


alexanderflink:
Suggested name @noaignite/centra-mock-data

Suggestions on imports:

```
import { products, selection } from '@noaignite/centra-mock-data'
```

maeertin:
suggestion: Perhaps even @noaignite/centra-mocks as the package name? Less to write ;) Additionally, we might include utils / generator functions for the mocks (not just data).

maeertin:
Initial package now created.

PR #179, opened 2024-08-21
feat(react-centra-checkout): add tests

### LINKED ISSUE #67: [react-centra-checkout] Add automated tests
image

Typescript files have been added to test coverage and now show up when running `yarn test:coverage` (or `yarn test` for all tests).

alexanderflink:
fixes #67 

This adds unit tests to react-centra-checkout using vitest and testing-library. The mock data is taken from Centras docs, so nothing sensitive there. Very much WIP.

TODO:
- [ ] refactor mock data into functions so we don't need a massive blob per mock
- [ ] add actually useful tests

### DIFF
--- a/packages/centra-mocks/tsup.config.ts
+++ b/packages/centra-mocks/tsup.config.ts
@@ -5,7 +5,7 @@ export default defineConfig({
   format: ['esm', 'cjs'],
   bundle: false, // TODO: Do we even need to bundle a mock package?
   clean: true,
-  dts: false, // TODO: Do we even need dts for a mock package?
+  dts: true,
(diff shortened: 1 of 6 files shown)


maeertin:
suggestion: What if we did a separate package for the centra mocks (eg. @noaignite/centra-mocks) that we also publish. That could be used for the tests here but also for our projects.

maeertin:
> suggestion: What if we did a separate package for the centra mocks (eg. @noaignite/centra-mocks) that we also publish. That could be used for the tests here but also for our projects.

Actually there's an open ticket for this #77 :)

adamsoderstrom (approved):
LGTM, @alexanderflink! 🌟

PR #181, opened 2024-08-21
feat(react-utils): add initial version of `createRequiredContext`

adamsoderstrom:


packages/react-utils/src/createRequiredContext.ts
```diff
@@ -0,0 +1,39 @@
+import * as React from 'react'
```
maeertin, line 1: todo: This import can be omitted with the current setup.

packages/react-utils/src/createRequiredContext.ts
```diff
@@ -0,0 +1,39 @@
...
+
+/**
+ * Create common context-related values from React with a required context.
```
maeertin, line 4: todo: I would try and rewrite this description somewhat as it's not really setting "context values", but rather pre-configures the context hook to require the Provider to be accessible, in turn ensuring the returned context value from the hook is always typed as intended.

packages/react-utils/src/createRequiredContext.ts
```diff
@@ -0,0 +1,39 @@
...
+ *
+ * @example
+ * ```typescript
+ * const [useFoo, FooConsumer, FooProvider] = createRequiredContext<FooContextValue, 'Foo'>('Foo')
```
maeertin, line 8: todo: Add the correct order of values within the tuple which currently do not match what is actually returned from the TS function.

packages/react-utils/src/createRequiredContext.ts
```diff
@@ -0,0 +1,39 @@
...
+ * ```
+ */
+
+export default function createRequiredContext<TContextProps, TName extends string>(
```
maeertin, line 12: todo: As this package `@noaignite/react-utils` is a new package we have already adopted the "no default exports" pattern. Should be a named export.

todo: The generic `TName` does not seem to be used anywhere, no need for it to be a generic with the current implementation.

packages/react-utils/src/createRequiredContext.ts
```diff
@@ -0,0 +1,39 @@
...
+  contextName: TName,
+) {
+  const context = React.createContext<null | TContextProps>(null)
+
+  if (process.env.NODE_ENV !== 'production') {
+    context.displayName = `${contextName}Context`
+  }
+
+  const providerName = `${contextName}Provider` as const
+  const hookName = `use${contextName}` as const
```
maeertin, line 22: nitpick: If these variables are only used for the error they might as well be defined within the if statements scope (`if (contextValue === null) {`).

suggestion: How would it look/feel if these were defined in a one-liner as below with only a single `as const`?
```ts
const [hookName, providerName] = [`use${contextName}`, `${contextName}Provider`] as const
```

maeertin (changes requested):
todo: Add the export of the new util to the barrel file.

review: Super nice helper, nice one @adamsoderstrom ✨  Left some review comments.

maeertin (approved):
LGTM ✨

PR #192, opened 2024-08-23
test(utils): convert all files to ts

maeertin:
Converted to TS and included some eslint/ts disables as we don't have to be as strict in test files.

No changeset needed.


adamsoderstrom (approved):
LGTM! 👍

PR #196, opened 2024-08-23
chore(prettier): convert config to esm

maeertin:


adamsoderstrom (changes requested):
LGTM!
Though, would like to test this in a project before merging.

maeertin:
> LGTM! Though, would like to test this in a project before merging.

@adamsoderstrom, you can test this directly in the accelerator repo. Just change `semi: false` to `semi: true` in `packages/prettier-config/index.mjs` and save a file in your desired editor of choice (assuming that you have the prettier plugin installed for your editor). Alternatively, you can run `pnpm format` to see if the prettier configuration has been set up correctly :)

adamsoderstrom (approved):
Realized that this package is not published! 🙃😄
Therefore, it should be safe to use, since it's only inside this repo.

Will approve! ⭐  @maeertin

PR #199, opened 2024-08-26
fix/react utils/bundle true

adamsoderstrom:
Prior to this commit, we've had an issue where directives like `"use
client"` gets stripped out of the `chunk-*` files created in the `dist`
folder, when specifying `bundle: true` in `tsup.config.ts`.
I tried to trace this issue and found a corresponding one in the `esbuild` repo (0).
In that, i found that @Seojunhwan had created a ESBuild plugin (1) that
resolves this issue. Though, as of this writing, the plugin doesn't
support single-quote directives, which we have in our repository. I
forked the repo and submitted a PR for that (2). Until then, i've
published my fix under my personal namespace and use it here.

[0]: https://github.com/evanw/esbuild/issues/3115
[1]: https://github.com/Seojunhwan/esbuild-plugin-preserve-directives
[2]: https://github.com/Seojunhwan/esbuild-plugin-preserve-directives/pull/15

### DIFF
--- a/packages/react-utils/tsup.config.ts
+++ b/packages/react-utils/tsup.config.ts
@@ -12,6 +12,7 @@ export default defineConfig({
   esbuildPlugins: [
     preserveDirectivesPlugin({
       directives: ['use client', 'use strict'],
+      // eslint-disable-next-line prefer-named-capture-group -- Is this a necessary rule?
       include: /\.(js|ts|jsx|tsx)$/,
       exclude: /node_modules/,
     }),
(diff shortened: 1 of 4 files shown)


maeertin (changes requested):
I have no current change requests but would want to test the built dist in an actual project. Marking as "request changes" to freeze merging temporarily :)

packages/react-utils/tsup.config.ts
```diff
@@ -12,6 +12,7 @@ export default defineConfig({
   esbuildPlugins: [
     preserveDirectivesPlugin({
       directives: ['use client', 'use strict'],
+      // eslint-disable-next-line prefer-named-capture-group -- Is this a necessary rule?
```
adamsoderstrom, line 15: Great question

maeertin: I'll leave the eslint-disable as is for now

maeertin (approved):
I've now tested this and it works perfectly. Awesome addition @adamsoderstrom

Issue #78, opened 2022-10-30
[create-app] Throw error message if yarn is not installed


maeertin:

adamsoderstrom:
Think we can close this, right, @maeertin ?

PR #200, opened 2024-08-26
feat(create-app): complete refactor meant to be used within a turborepo project

maeertin:
Scaffold NoA Ignite templates within a turborepo project.


maeertin:
Now compatible with all the bigger package managers #78

adamsoderstrom (commented):
Maybe we can add a `--no-turbo` flag to make `@noaignite/create-app` decoupled from Turborepo?

packages/create-app/src/index.ts
```diff
@@ -0,0 +1,410 @@
...
+  console.log()
+}
+
+function formatTargetDir(targetDir: string | undefined) {
+  return targetDir?.trim().replace(/\/+$/g, '')
+}
+
+function formatCommand(command: string) {
+  const pkgInfo = pkgFromUserAgent(process.env.npm_config_user_agent)
+  const pkgManager = pkgInfo ? pkgInfo.name : 'npm'
```
adamsoderstrom, line 301: **question:** How come we fallback to `npm` here?

Have a hunch that we may provide better UX if we can print an error if we can't properly determine the package manager.

Maybe we can add the functionality to pass the package manager like turborepo (`-m`) as well?

packages/create-app/src/index.ts
```diff
@@ -0,0 +1,410 @@
...
+    const formattedCommand = formatCommand(customCommand)
+    const [command, ...args] = splitCommand(formattedCommand) as [string, ...string[]]
+    spawnSync(command, args, { stdio: 'inherit' })
+  }
+
+  // Step 4. Install dependencies
+
+  console.log(`\nDone. Now installing dependencies.\n`)
+
+  if (pkgManager) {
```
adamsoderstrom, line 280: **chore:** Make the `pkgManager` type `string` and not `string | undefined`

packages/create-app/src/index.ts
maeertin: A lot from this PR was inspired by `vite-create` and that is how they handle checking for what package manager you used to run the executable. That said, the fallback is for those package managers that do not have support for  `npm_config_user_agent`. This is supported by the major package managers like `bun`, `npm`, `pnpm` & `yarn`.

As for the `-m` flag in turbo, this script handles it differently. The package manager you used at install time is the one that will be used internally for installation.

packages/create-app/src/index.ts
maeertin: Updated :)

PR #216, opened 2024-08-29
fix(react-utils): wrap provider & consumer to register them as client components

maeertin:
This is a fix for the nextJs app router


adamsoderstrom (approved):
Good catch, @maeertin!

Looks good to me! 😊

PR #225, opened 2024-09-11
feat: eslint files as cjs & updated --ignore-pattern for lint scripts

maeertin:
Converting these files to `.cjs` automatically sets the environment to node getting rid of the `File is a CommonJS module; it may be converted to an ES6 module.` warning.


adamsoderstrom (approved):
Interesting!
Looks good!

PR #226, opened 2024-09-13
react-centra-checkout: add various tests

adamsoderstrom:


packages/react-centra-checkout/package.json
```diff
@@ -31,8 +31,10 @@
     "build": "tsup",
     "build:docs": "typedoc",
     "lint": "eslint src/ --ignore-pattern '**/*.test.*'",
-    "test-coverage": "vitest run",
-    "test-types": "tsc --noEmit"
+    "test-types": "tsc --noEmit",
+    "test:unit": "vitest run",
+    "test:unit:coverage": "vitest run --coverage",
+    "test:unit:watch": "vitest"
```
maeertin, line 37: Nice initiative @adamsoderstrom but let's discuss a standard for our script naming convention; Either use hyphen or semicolon? And also what the different tests should be called as these renamed tests would not be executed currently from the root "turbo test" script.

maeertin (changes requested):
Nice initiative @adamsoderstrom but let's discuss a standard for our script naming convention; Either use hyphen or semicolon? And also what the different test types should be called as these renamed tests would not be executed currently from the root "turbo test" script.

maeertin (approved):
LGTM

PR #227, opened 2024-09-13
 Expose `calculateContrast`, `calculateLuminance` and `hexToRGB` utility functions

### LINKED ISSUE #205: Export internal helpers found in `utils/colorContrast`
Currently there are three internal helpers found in the [utils/colorContrast](https://github.com/noaignite/accelerator/blob/main/packages/utils/src/colorContrast.ts). These are `calculateLuminance`, `calculateContrast` and `hexToRgb`. Would be nice if these were moved to their own files and exported as public utils.

adamsoderstrom:
- **feat(utils): expose `calculateLuminance` to consumer of package**
- **feat(utils): expose `calculateContrast` to consumer of package**
- **feat(utils): expose `hexToRGB` to consumer of package**
- **docs(changeset): Expose `calculateContrast`, `calculateLuminance` and `hexToRGB` utility functions**

fixes #205


maeertin:
Nice job @adamsoderstrom :) Do we feel the need to write tests for these functions or is it good enough that `colorContrast` which uses them all has a test?

packages/utils/src/colorContrast.ts
Dashice, line 1:
```
  // Determine the best fallbackColor, being black or white
  const luminance = calculateLuminance(baseRGB)
  absoluteContrast = luminance > 0.5 ? '#000' : '#fff'
```

This part of the code is not correct (my bad). What we are looking for is:

```
  // Calculate contrast ratios with black and white text
  const blackContrast = calculateContrast(baseRGB, [0, 0, 0]);
  const whiteContrast = calculateContrast(baseRGB, [255, 255, 255]);

  // Determine the best absolute contrast color
  const absoluteContrast = blackContrast > whiteContrast ? '#000' : '#fff';
```

maeertin: Please provide PR 🥳

adamsoderstrom:
> Nice job @adamsoderstrom :) Do we feel the need to write tests for these functions or is it good enough that `colorContrast` which uses them all has its test?

I do think that we shouldn't implicitly rely on tests.

I could write the tests according to the examples! 😊

packages/utils/src/colorContrast.ts
adamsoderstrom: Fixed it in this PR!

Dashice (approved):
Tested against the colors that were problematic; this seems to work.

PR #230, opened 2024-09-16
feat(react-utils): add useId hook

Dashice:


maeertin:
I'm not too fond of adding this hook for a few reasons, that being:
* Modifies Reacts generated id to make it queryable with `getElementById` which alters/disregards Reacts design choice for this decision.
* Reacts unmodified id is still queryable for example via `querySelector('[id="${reactId}"]')`.
* Adding a prefix/suffix to a string (the id in this case) is easily done with template literals.
* Has the same name as Reacts own `useId` hook giving the impression of this hook being a patched version.

If we do feel the need for a `useId` hook with functionality to prefix/suffix I would suggest going for a hook that is called just that. For example `usePrefixedId`.

What are your thoughts @adamsoderstrom @Dashice ?

Dashice:
Closed due to unpopularity.

PR #242, opened 2024-09-19
build: upgrade `esbuild-plugin-preserve-directives`

adamsoderstrom:


adamsoderstrom:
Considering that no changeset is needed, since it shouldn't changes any published code.

maeertin (approved):
LGTM ✨

PR #244, opened 2024-09-20
docs(utils): expand on documentation

maeertin:


packages/utils/src/range.ts
```diff
@@ -1,11 +1,17 @@
...
- * @param step - The step between each number
- * @param inclusive - Include the stop number in the range
- * @returns An array of numbers
+ * @param start - The starting value of the range.
+ * @param stop - The ending value of the range.
+ * @param step - The increment between each number in the generated range. Can
+ * be a positive or negative number.
+ * @param inclusive - If `true`, the stopping value will be included in the
+ * generated array. If `false`, the stopping value will be excluded.
+ * @returns An array of numbers representing the specified range.
```
adamsoderstrom, line 14: **todo:** keep the same terminology as in the description of the function.

Either replace `stopping point` with `value` or vice versa.

packages/utils/src/pick.ts
```diff
@@ -1,9 +1,14 @@
 /**
- * Pick keys from object.
+ * The `pick` function creates a new object by selecting specified keys from
+ * the original object. It takes an object and an array of keys to include,
+ * returning a new object that contains only the properties specified in the
+ * keys array. This is useful for extracting relevant data from an object while
+ * ignoring other properties.
```
adamsoderstrom, line 6: **thought:** Don't really know if we need to express why the function is useful.

packages/utils/src/pick.ts
```diff
@@ -1,9 +1,14 @@
 /**
- * Pick keys from object.
+ * The `pick` function creates a new object by selecting specified keys from
+ * the original object. It takes an object and an array of keys to include,
+ * returning a new object that contains only the properties specified in the
```
adamsoderstrom, line 4: **thought:** should we wrap parameter names in template strings in the description aswell?

packages/utils/src/pick.ts
```diff
@@ -1,9 +1,14 @@
...
+ * keys array. This is useful for extracting relevant data from an object while
+ * ignoring other properties.
  *
- * @param object - Object to pick keys from.
- * @param keys - Array of keys to pick from object.
- * @returns Object with picked keys.
+ * @param object - The original object from which keys will be selected.
+ * @param keys - An array of keys (property names) to include in the new object.
```
adamsoderstrom, line 9: **praise:** good w/ clarification (property names)!

packages/utils/src/omit.ts
```diff
@@ -1,9 +1,15 @@
 /**
- * Omit keys from object.
+ * The `omit` function creates a new object by omitting specified keys from the
+ * original object. It takes an object and an array of keys to exclude,
+ * returning a new object that contains only the properties not specified in
+ * the keys array. This is useful for filtering out sensitive or unnecessary
+ * information from an object before using it.
```
adamsoderstrom, line 6: **thought:** I don't really know if we _need_ to present verbally what the function does

I think the example perfectly shows this.

packages/utils/src/normalize.ts
```diff
@@ -1,15 +1,25 @@
...
+ * range. This is useful for standardizing data, making it easier to work with
+ * values across different scales.
  *
- * @param value - The value to normalize
- * @param min - The minimum value of the range
- * @param max - The maximum value of the range
- * @returns The normalized value
+ * @param value - The number to be normalized within the range.
+ * @param min - The minimum value of the range.
+ * @param max - The maximum value of the range.
```
adamsoderstrom, line 10: **thought:** why not "number" here as well?

packages/utils/src/normalize.ts
```diff
@@ -1,15 +1,25 @@
...
- * normalize(5, 0, 10) // 0.5
- * normalize(5, 0, 5) // 1
+ * // Normalize a value within the range [0, 100]
+ * normalize(50, 0, 100)); // 0.5
+ * // Normalize a value in a custom range [10, 20]
+ * normalize(15, 10, 20)); // 0.5
+ * // Normalize a value below the minimum (the function doesn't clamp)
+ * normalize(5, 10, 20)); // -0.5
+ * // Normalize a value above the maximum
+ * normalize(25, 10, 20)); // 1.5
```
adamsoderstrom, line 22: **nitpick:** Write "normalize a number" instead

packages/utils/src/lerp.ts
```diff
@@ -1,15 +1,24 @@
 /**
- * Linear interpolation between two numbers.
+ * The `lerp` (linear interpolation) function calculates a value between a
```
adamsoderstrom, line 2: **thought:** rename the function to `interpolateLinearly`

packages/utils/src/lerp.ts
```diff
@@ -1,15 +1,24 @@
...
+ * returns the start value, `1` returns the end value, and values between 0
+ * and 1 return points in between. This is commonly used in animation,
+ * movement, or smoothing transitions.
  *
- * @param start - The start value.
- * @param end - The end value.
- * @param amount - The amount to interpolate between the two values.
- * @returns The interpolated value.
+ * @param start - The starting value of the interpolation.
+ * @param end - The ending value of the interpolation.
```
adamsoderstrom, line 10: **nitpick:** use "number" instead of "value"

adamsoderstrom (changes requested):

packages/utils/src/omit.ts
maeertin: Useful to have when consuming this information via the documentation site.

packages/utils/src/lerp.ts
maeertin: Went originally with `lerp` as it's a standard name for this function.

adamsoderstrom (approved):

packages/utils/src/normalize.ts
maeertin: Updated copy to `value` as this text is not directly tied to it's ts type.

packages/utils/src/normalize.ts
maeertin: Updated copy elsewhere to `value` as this text is a descriptive text and not directly tied to it's ts type.

PR #245, opened 2024-09-21
chore(generateMarkdown): convert to JS

adamsoderstrom:
Due to our setup uses Node v20, we can't use the
`--experimental-strip-types` flag yet.


maeertin:
Thanks for the contribution ✨ But I solved it with [ts-node](https://www.npmjs.com/package/ts-node) instrad to keep typechecking :)

PR #248, opened 2024-09-24
docs(react-centra-checkout): correct headings in `CHANGELOG.md`

adamsoderstrom:


packages/react-centra-checkout/CHANGELOG.md
```diff
@@ -7,17 +7,17 @@
...
 
   `resetSelection` now returns a `Promise<void>` instead of `void`, which allows for better handling of potential race conditions.
 
-  ## WHY
+  ##### WHY
 
   To evade potential race conditions between different state changes to `selection`.
 
-  ## HOW
+  ##### HOW
```
maeertin, line 20: One can ofc write CHANGELOGs as one pleases but I do love standards and following conventions. With that said, the convention seems to be to not add headlines to not disturb the general flow of headline hierarchy generated by the CHANGELOG scripts.
Additionally, perhaps that having all this information (What/Why/How) is too much for the CHANGELOG and is meant to be contained within the commit message. Thinking about it now, I find it makes sense as every CHANGELOG entry links to it's respective commit.

References:
* https://github.com/vercel/vercel/blob/main/packages/next/CHANGELOG.md#400
* https://github.com/mui/material-ui/blob/master/CHANGELOG.md#600
* https://github.com/tailwindlabs/headlessui/blob/main/packages/@headlessui-react/CHANGELOG.md#200---2024-05-06

The only thing missing for this to work is for the created changesets to be part of same commit as the implemented logic. For this we need to update the changeset config to not commit. Created an issue for this #249

adamsoderstrom: I've now removed the "Breaking change" section completely, @maeertin

maeertin (approved):
LGTM

Issue #249, opened 2024-09-25
[changeset] Update changeset config to not commit


maeertin:
Update changeset config to not commit. The idea is for the created changeset to be part of the same commit that implements the feature so that the linked commit in the CHANGELOG actually has all the related code.

Ideally we could update the CONTRIBUTING.md to describe this flow.

PR #252, opened 2024-09-25
chore(changeset): changeset no longer commits for you + add changeset section to CONTRIBUTING.md

maeertin:
As stated in ticket #249. Update changeset config to not commit. The idea is for the created changeset to be part of the same commit that implements the feature so that the linked commit in the CHANGELOG actually has all the related code.


adamsoderstrom (approved):
LGTM!

PR #257, opened 2024-09-26
docs: git ignore generate docs file + format generated page titles

adamsoderstrom:
By doing so, we could avoid git-tracking the generated files.
Git tracking the generated files could result in human error by manually making edits to these files, while in other commit generating new docs.

Not git tracking these files removes this possibility for human error.


maeertin (approved):
LGTM

PR #264, opened 2024-09-27
build(typescript-config): update module to Preserve & moduleResolution to Bundler

maeertin:
The current `typescript-config` template from `create-app` is not correctly configured to work with our packages as it is. You get this error `The current file is a CommonJS module whose imports will produce 'require' calls`. Tweaking the tsconfig to the following solves this issue:

```json
"module": "Preserve",
"moduleResolution": "Bundler",
```

I suppose the `typescript-config` template from `create-app` and the accelerator's `typescript-config` might come to diverge at some point but perhaps that is not it. Hence, I updated the accelerators own tsconfig with the same tweak.

This still being in line with Matt Pocock's latest post regarding tsconfig:
```json
// Every other "module" option is wrong
"module": "NodeNext", // (or "Preserve")
```

All test & build scripts pass.


adamsoderstrom (approved):
I agree to your reasoning.

I'm not super experienced to configuring TypeScript, but drawing influences from Matt seems liable. We could also see it in the following manner:

> Why fix something that ain't broken?

It has now become apparent that it was and therefore we fix it.

The same reasoning was applied when we made the initial versions of the configs and could be applied if future bugs would occur.

Sweet! 💥

PR #265, opened 2024-09-27
[react-centra-checkout]: remove nested directories

adamsoderstrom:
By doing so, we align the structure that we have in other packages in the repository, while still keeping the export paths intact.

The directories hasn't provided any value in the package.
If we, for some reason, would need directories again, we can always add it back again.


maeertin (approved):
LGTM ✨

PR #266, opened 2024-09-27
ci(changeset): ensure existence of changeset in PR

adamsoderstrom:
Tested locally by applying the changes on an ongoing PR (fails, due to lacking changeset, as it should).
I also tested it how it performs on the `main` branch by applying temporarily there as well (doesn't fail, due to being `main`).

Adding the `changeset status` call to be last in the script improves DX by seeing that the CI pipeline passes before you've necessarily submitted a changeset.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -16,6 +16,9 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v4
+        with:
+          ref: ${{github.event.pull_request.head.sha}}
+          fetch-depth: 0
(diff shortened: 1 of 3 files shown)


adamsoderstrom:
One could also see the checks are passing on this branch in the current CI (See the "Run pnpm run ci:test" step), even though it doesn't contain any changeset.
As i interpret it, it's because we haven't made any changes to a package in this PR and therefore there's no need for versioning them (see, https://github.com/changesets/changesets/blob/main/packages/cli/README.md#status).

maeertin (approved):
This is great, ensures a choice of adding a changeset or not. Cheers ✨

PR #283, opened 2024-10-10
feat(react-utils): add new globals option to createRenderBlock

maeertin:


packages/react-utils/src/createRenderBlock.tsx
```diff
@@ -34,25 +35,33 @@ import { ErrorBoundary, type ErrorBoundaryProps } from './ErrorBoundary'
  * }
  *
  * const adapters = {
- *   Hero: async (props: HeroData['props']): Promise<HeroProps> => {
+ *   Hero: async (props: HeroData['props'], _, globals): Promise<HeroProps> => {
```
adamsoderstrom, line 38: **suggestion:** rename `_` to `additionalData` and incorporate example

For clarifying the mapping passed properties and where they apply.

adamsoderstrom (changes requested):

adamsoderstrom (approved):
Sweet! 🙌

PR #289, opened 2024-10-14
feat(react-utils): add useElementSize hook

Dashice:


adamsoderstrom (approved):
Despite not trying it out personally, i picture that the type narrowing will work exactly like `useWindowSize` (#290), which is great!

Sweet! 🙌

maeertin (approved):

PR #302, opened 2024-10-18
feat(eslint-config): allow dev deps in tsup config

maeertin:
Reason:
eslint error wanting `tsup` to be in deps and not devDeps. Any other way to solve this?

Changes:
* Allow dev deps such as `tsup` in tsup config.
* `'prefer-named-capture-group': 'off', // Too strict.` line was added to `template-eslint-config` to stay in parity with the `eslint-config` used by the monorepo.


packages/create-app/template-eslint-config/library.cjs
```diff
@@ -58,5 +60,6 @@ module.exports = {
 
     'lines-around-directive': ['error', 'always'], // Enable & let prettier format.
     'no-console': ['error', { allow: ['info', 'warn', 'error'] }], // Allow warn and error logs.
+    'prefer-named-capture-group': 'off', // Too strict.
```
adamsoderstrom, line 63: Agree!

adamsoderstrom (approved):

PR #305, opened 2024-10-18
feat(react-utils): pass blockType to block component

maeertin:


maeertin:
Close in favor of https://github.com/noaignite/accelerator/pull/306

PR #306, opened 2024-10-21
feat(react-utils): pass `blockType` to block, `context` always as object & export new types

maeertin:
- `blockType` is now passed to block component.
- Adapter argument `additionalData` has been renamed to `context`
- Adapter argument `context` is now always an object containing `index` & additional keys one provides to the renderer.
- Export new `BlockTypeMap` & `BlockAdapter` types.


adamsoderstrom (approved):
Interesting!

PR #319, opened 2024-10-28
feat: add initial version of `createGetCentraWebhookEvents`

adamsoderstrom:


packages/react-centra-checkout/package.json
```diff
@@ -37,7 +37,8 @@
   },
   "dependencies": {
     "@noaignite/utils": "workspace:*",
-    "js-cookie": "^3.0.1"
+    "js-cookie": "^3.0.1",
+    "zod": "^3.23.8"
```
adamsoderstrom, line 41: Very unsure if we want to add `zod` as dependency in this package.
Considered more fine, if we can assure that it would be properly tree-shaken away from clientside usage of the package.

Dashice: This seems redundant. It is used in one place; I'd rather you create predicates for `bodyParserSchema` or if the intention is to use it package-wide; then I'd introduce Zod if it serves a more concrete purpose.

packages/react-centra-checkout/src/server/createGetCentraWebhookEvents.ts
```diff
@@ -0,0 +1,166 @@
...
+
+  return hmac.digest('hex')
+}
+
+/**
+ * Parse a named parameters string into an object.
+ * @example
+ * parseNamedParametersString('foo=bar,baz=qux') // \{ foo: 'bar', baz: 'qux' \}
+ */
+const parseNamedParametersString = (input: string): Record<string, string> => {
```
Dashice, line 18: Keep in mind for the foreseeable future; `objectFromURLSearchParams` PR may solve this similarly. Consider seeing if it can be applicable here for code-reuse.

packages/react-centra-checkout/src/server/createGetCentraWebhookEvents.ts
```diff
@@ -0,0 +1,166 @@
...
+    getRawBody: (request: TRequest) => unknown
+  }
+  secret?: string
+}
+
+/**
+ * Create get Centra webhook events.
+ * Parses and returns the events passed from Centra.
+ *
+ * @see https://centra.dev/docs/services/centra-webhooks
```
maeertin, line 64: suggestion(non-blocking): Move `@see` block as the last "tag" to be consistent in the project as I think this also decides the order in the generated docs.

packages/react-centra-checkout/src/server/createGetCentraWebhookEvents.ts
```diff
@@ -0,0 +1,166 @@
...
+ *     return Response.json({ message: 'Error' }, { status: 500 })
+ *   }
+ *
+ *   if ('products' in results) {
+ *     return Response.json({ products: results.products }, { status: 200 })
+ *   }
+ * }
+ * ```
+ */
+export function createGetCentraWebhookEvents<TRequest>(
```
Dashice, line 85: Consider an explicit return type here to avoid an excess of `as const` assertions. Changes to this function could inadvertently change the return type without somebody noticing.

Dashice: E.g `export function createGetCentraWebhookEvents<TRequest>(config: CreateGetCentraWebhookEventsConfig<TRequest> ): ResponseType { ... }`

packages/react-centra-checkout/src/server/createGetCentraWebhookEvents.ts
```diff
@@ -0,0 +1,166 @@
...
+      }
+    }
+
+    const payloadParsed = JSON.parse(body.payload) as Partial<Record<EventTypes, string[]>>
+
+    return [undefined, payloadParsed] as const
+  }
+}
+
+export const nextAppRouterAdapter: CreateGetCentraWebhookEventsConfig<Request>['adapter'] = {
```
maeertin, line 137: question: Are we forced to be next specific here? If wanting to support other frameworks how do we go about it? Just food for though :)

adamsoderstrom: No, as i've structured it now, the consumer is expected to import `createGetCentraWebhookEvents` and either:
- Import a predefined adapter (`nextAppRouterAdapter`, `nextPagesRouterAdapter`, for now).
- Create one on a project-level and pass it to `createGetCentraWebhookEvents`.

packages/react-centra-checkout/src/server/createGetCentraWebhookEvents.ts
adamsoderstrom: Noted!
I tried if i could use `objectFromUrlSearchParams` and see if it "just works", which, unfortunately, it doesn't.
Would be nice to get a shared solution, though, since the only thing that differs interface-wise is a separator.

packages/react-centra-checkout/src/server/createGetCentraWebhookEvents.ts
```diff
@@ -0,0 +1,163 @@
...
+
+const parseRawBody = (rawBody: unknown): { payload: string } | null => {
+  if (rawBody && isObject(rawBody) && 'payload' in rawBody && typeof rawBody.payload === 'string') {
+    return rawBody as { payload: string }
+  }
+
+  return null
+}
+
+type CreateGetCentraWebhookEventsConfig<TRequest> = {
```
maeertin, line 38: todo: Export type

packages/react-centra-checkout/src/server/createGetCentraWebhookEvents.ts
```diff
@@ -0,0 +1,163 @@
...
+  adapter: {
+    isRequestMethodPost?: (request: TRequest) => boolean
+    getHeader: (request: TRequest, headerKey: string) => string | null | undefined
+    getRawBody: (request: TRequest) => unknown
+  }
+  secret?: string
+}
+
+type CentraWebhookEventsError =
```
maeertin, line 47: todo: Export type
suggestion: Add ts-docs

packages/react-centra-checkout/src/server/createGetCentraWebhookEvents.ts
```diff
@@ -0,0 +1,163 @@
...
+  | [{ message: 'Invalid request method' }]
+  | [{ message: 'No signature' }]
+  | [{ message: 'Invalid request body' }]
+  | [{ message: 'Invalid signature' }]
+type CentraWebhookEventsData = [undefined, Partial<Record<Events, string[]>>]
```
maeertin, line 52: todo: Export type
suggestion: Add ts-docs

packages/react-centra-checkout/src/server/createGetCentraWebhookEvents.ts
```diff
@@ -0,0 +1,163 @@
...
+  return null
+}
+
+type CreateGetCentraWebhookEventsConfig<TRequest> = {
+  adapter: {
+    isRequestMethodPost?: (request: TRequest) => boolean
+    getHeader: (request: TRequest, headerKey: string) => string | null | undefined
+    getRawBody: (request: TRequest) => unknown
+  }
+  secret?: string
```
maeertin, line 44: suggestion: Require `secret` property, e.g. `secret: string | null`

packages/react-centra-checkout/src/server/createGetCentraWebhookEvents.ts
```diff
@@ -0,0 +1,163 @@
...
+ * Create get Centra webhook events.
+ * Parses and returns the events passed from Centra.
+ *
+ * @example
+ * ```typescript
+ * const getCentraWebhookEvents = createGetCentraWebhookEvents({
+ *   adapter: nextAppRouterAdapter
+ * })
+ *
+ * const POST = async (req: Request) => {
```
adamsoderstrom, line 66:
```suggestion
 * const POST = async (request: Request) => {
```

packages/next-centra-checkout/src/createGetCentraWebhookEvents.ts
```diff
@@ -0,0 +1,224 @@
...
+    if (Array.isArray(header)) {
+      throw new Error('Multiple headers with same key passed.')
+    }
+
+    return header
+  },
+  getRawBody: (req) => {
+    return req.body as unknown
+  },
+} satisfies CreateGetCentraWebhookEventsConfig<NextApiRequest>['adapter']
```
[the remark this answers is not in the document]
adamsoderstrom: I do enjoy the `satisfies` expression more.

PR #322, opened 2024-11-04
feat(utils): add normalizeToArray helper

maeertin:


Dashice (approved):

packages/utils/src/normalizeToArray.ts
```diff
@@ -0,0 +1,22 @@
...
+ *
+ * @example
+ * ```ts
+ * normalizeToArray(5); // [5]
+ * normalizeToArray('foo'); // ['foo']
+ * normalizeToArray([1, false]); // [1, false]
+ * normalizeToArray(undefined); // []
+ * ```
+ */
+export function normalizeToArray<T>(value: T | T[] | undefined): T[] {
```
Dashice, line 19: Passing `NaN`, `0`, `null` returns `[]`, is this intentional in regards to your vision for this functions behavior.

Dashice: Is it also intentional that passing `[undefined]` returns itself`

maeertin: Good catch, updated the PR 👍

PR #323, opened 2024-11-04
feat(utils): add random helper

maeertin:


packages/utils/src/random.ts
```diff
@@ -0,0 +1,20 @@
+/**
```
Dashice, line 1: Maybe you want to mention whether the value is excluding / including the passed arguments?

Dashice (approved):

PR #324, opened 2024-11-04
feat(utils): add `toggleArrayValue` helper

maeertin:


packages/utils/src/toggle.ts
```diff
@@ -0,0 +1,18 @@
...
+ * present.
+ *
+ * @example
+ * ```ts
+ * toggle([1, 'foo', true], 1); // ['foo', true]
+ * toggle([1, 'foo', true], true); // [1, 'foo']
+ * toggle([1, 'foo', true], 'baz'); // [1, 'foo', true, 'baz']
+ * ```
+ */
+export function toggle<T>(array: T[], value: T): T[] {
```
Dashice, line 16: - Should this be called `toggle`.
- Currently, item(s) previously toggled do not retain their index, should they?
- Should this work for Objects / Sets / Maps, how?
- Without passing references, you cannot toggle reference types; should you be able to?

maeertin: > * Should this be called toggle?

It could be called something like `toggleArrayValue` instead. If so, there are some questions that come to mind. What are we reserving the name `toggle` for? And what is the difference between this naming versus for example the function `omit` which handles objects and is not called something like `omitObjectKey`?

> * Currently, item(s) previously toggled do not retain their index, should they?

I would prefer to not complicate this helper as it's current intended purpose is very straightforward. All it does is to toggle the existence of `value` from the `array` in it's current state passed as the first argument. A very pure function.

> * Should this work for Objects / Sets / Maps, how?

Perhaps that it would make sense for a `Set`, but not for `Object`s or `Map`s. Object-like structures already have their own dedicated helpers like `omit` and `pick`. I would rather not go the route of expanding `omit` to also receive arrays.

> * Without passing references, you cannot toggle reference types; should you be able to?

This could be a good addition. Although, I would rather see this helper added to the repo in it's simplest form as a first version. Adding this functionality thereafter would give us a very clear comparison if we think the added amount of code is worth it.

Dashice: Change name to `toggleArrayValue`... Consider future implementation of a `toggle` super func that allows other array-like types to be passed. Approved.

Dashice (changes requested):
https://github.com/noaignite/accelerator/pull/324/files#r1957941429

packages/utils/src/toggle.ts
maeertin: Sounds good, going ahead with this :)

adamsoderstrom (approved):

PR #334, opened 2024-11-13
fix(prettier): update "faulty" config & format

maeertin:
A quick investigation leads me to believe that this might only be an issue if one is running the latest node.


Dashice (approved):
LGTM

PR #348, opened 2024-11-20
test(react-utils): add test for useWindowSize

Dashice:


packages/react-utils/src/useWindowSize.test.ts
```diff
@@ -0,0 +1,51 @@
...
+  window.innerWidth = 800
+  window.innerHeight = 600
+  window.dispatchEvent(new Event('resize'))
+})
+
+const dispatch = (width: number, height: number) => {
+  window.innerWidth = width
+  window.innerHeight = height
+  window.dispatchEvent(new Event('resize'))
+}
```
maeertin, line 15: suggestion: Could shorten by defining `dispatch` first and then using within `beforeEach`:

```ts
const dispatch = (width: number, height: number) => {
  window.innerWidth = width
  window.innerHeight = height
  window.dispatchEvent(new Event('resize'))
}

beforeEach(() => {
  dispatch(800, 600)
})
```

packages/react-utils/src/useWindowSize.test.ts
```diff
@@ -0,0 +1,51 @@
+import { act, renderHook } from '@testing-library/react'
+import { beforeEach, expect, test } from 'vitest'
+import { useWindowSize } from './useWindowSize'
+
+beforeEach(() => {
```
maeertin, line 5: question: Are we deliberately skipping the `describe` block for the new tests? Could be lack of knowledge but my initial worry was that life cycle methods like `beforeEach` might be registered to a global scope and might cause side-effects for other tests. Is this perhaps not the case? Feels like I usually I see this pattern:
```ts
describe('myFunction', () => {
  beforeEach(() => {
    // ...
  })

  it('should do...', () => {
    expect(...)
  })
})
```
For example https://github.com/vercel/next.js/blob/cf492c23e569ac32a9412d284b04f8538ef2d64c/packages/next/src/shared/lib/router/adapters.test.tsx#L5

maeertin: Just using the `test` function looks more than good enough 👍

maeertin (approved):

PR #356, opened 2024-11-21
feat(utils): add formatDate

Dashice:


adamsoderstrom (commented):
I get a feeling that we've only built a wrapper around the Date API with default options.
I do think that there are some valuable knowledge added in this PR and helps us all do better work, but i wonder if one would prefer using the Date API directly in the individual project and then rely on only one abstraction.

Maybe a scaffolding of this function would be something?

maeertin:
> I get a feeling that we've only built a wrapper around the Date API with default options. I do think that there are some valuable knowledge added in this PR and helps us all do better work, but i wonder if one would prefer using the Date API directly in the individual project and then rely on only one abstraction.
> 
> Maybe a scaffolding of this function would be something?

I also somehow feel that the implementation in it's current form for the most part simply sets opinionated defaults and not much more. Or is there something I'm not seeing?

Dashice:
Closed due to inactivity.

PR #360, opened 2024-11-25
feat(utils): add collate

Dashice:


packages/utils/src/collate.ts
```diff
@@ -0,0 +1,26 @@
...
+export type CollateOptions = DistributiveOmit<Intl.CollatorOptions, 'usage'>
+
+/**
+ * Sorts a list of strings alphabetically using the given locales and options.
+ *
+ * @param list - The list of `string` values to sort.
+ * @param locales - The locales to use for sorting.
+ * @param options - {@link CollateOptions} configurable options.
+ *
+ * @returns A new list of strings sorted alphabetically.
```
adamsoderstrom, line 12: I think we should use `toSorted` to ensure that this statement is true.
From what i've heard, `sort` will modify the current instance of `list`.

adamsoderstrom (changes requested):

packages/utils/src/collate.ts
Dashice: Nice catch. Geez, what a fail on my part!

packages/utils/src/collate.ts
```diff
@@ -0,0 +1,26 @@
...
+ *
+ * @example
+ * ```ts
+ * const list = ['B', 'A', 'a', 'Z', 'ø', 'æ', 'ä']
+ *
+ * collate(list, 'de-DE')
```
Dashice, line 18: Missing result.

packages/utils/src/collate.ts
```diff
@@ -0,0 +1,26 @@
...
+ * collate(list, 'nb-NO', { caseFirst: 'upper' })
```
Dashice, line 19: Missing result.

Dashice:
Closed. Replaced with #399

PR #366, opened 2024-11-27
fix(react-utils): fix threshold reference -> primitive dep

Dashice:
This [post](https://github.com/facebook/react/issues/14476#issuecomment-471199055) explains the approach used here and was written by Dan, a React Core Team Member.


adamsoderstrom (approved):
Was curious if one could use, `Array.join`, but feels a bit overkill. :)
Looks good to me!

PR #367, opened 2024-11-27
feat(react-utils): add useDismiss

Dashice:


packages/react-utils/src/useDismiss.test.tsx
```diff
@@ -0,0 +1,152 @@
...
+
+function Scenario3() {
+  const outerRef = useRef<HTMLDivElement>(null)
+  const innerRef = useRef<HTMLDivElement>(null)
+
+  const [innerOpen, setInnerOpen] = useState(true)
+  const [outerOpen, setOuterOpen] = useState(true)
+
+  useDismiss(outerRef, setOuterOpen.bind(null, false), { when: outerOpen })
+  useDismiss(innerRef, setInnerOpen.bind(null, false), { when: innerOpen })
```
adamsoderstrom, line 120: **praise:** So nice with the `.bind`!

packages/react-utils/src/useDismiss.ts
```diff
@@ -0,0 +1,112 @@
...
+    (event: PointerEvent) => {
+      const element = ref.current
+      const target = event.target
+
+      if (!(element instanceof Element) || !(target instanceof Element)) return
+      if (Array.isArray(pointer) && !pointer.includes(event.pointerType)) return
+      if (typeof pointer === 'string' && event.pointerType !== pointer) return
+      if (element.closest('[inert]') ?? element.contains(target)) return
+      if (element.querySelector('[data-dismissible]')) return
+      if (event.button !== 0) return
```
adamsoderstrom, line 89: **suggestion:** Add `console.warn` on suitable scenarios

I do think that it would be nice, in terms of debugging. For example:
```
  if (!(element instanceof Element) || !(target instanceof Element)) return
```

Would be great to get some hints that the `element` that the developer has passed is not valid.

packages/react-utils/src/useDismiss.test.tsx
adamsoderstrom, line 1: As Martin asked in this [review](https://github.com/noaignite/accelerator/pull/348/files#r1916881377), i think we should be consistent with the `describe` function. I think we could either remove it throughout the project, or use it in this test.

maeertin: Although looking at the vitest output, I do think just using `test` is much cleaner 🤭 Should we look at this later? Maybe there's good uses cases for them both.

packages/react-utils/src/useDismiss.test.tsx
maeertin: question: How come `.bind` is needed? :o

packages/react-utils/src/useDismiss.test.tsx
adamsoderstrom: Sure, i don't really care or have an opinion! 😄 
But i do think that's out of scope for this PR and just keep it consistent with the current setup.
Or maybe we'll simply put this PR on hold and refactor the tests throughout the code base?

maeertin: What I tried to convey was that I think there might be cases where one or the other might give a cleaner output. Using describe+it vs test where it suits the test :)

adamsoderstrom: Yeah, agreed! 👍

adamsoderstrom (approved):

packages/react-utils/src/useDismiss.ts
maeertin: This could be nice, what do you think @Dashice ?

packages/react-utils/src/useDismiss.ts
```diff
@@ -0,0 +1,112 @@
...
+ *
+ * return (
+ *   <>
+ *     <button type="button" onClick={() => setIsOpen(true)}>Open</button>
+ *     {isOpen && (<div ref={ref}>Content</div>)}
+ *   </>
+ * )
+ * ```
+ */
+export const useDismiss = (
```
maeertin, line 57: question: I'm wondering if a name like "click away/outside" might be more recognisable? Looking at implementations like:
* https://github.com/juliencrn/usehooks-ts/blob/master/packages/usehooks-ts/src/useOnClickOutside/useOnClickOutside.ts
* https://github.com/mui/material-ui/blob/master/packages/mui-material/src/ClickAwayListener/ClickAwayListener.tsx

packages/react-utils/src/useDismiss.test.tsx
Dashice: Cleanliness. Wanted both on same line. Otherwise prettier will turn these into 6 lines total.

packages/react-utils/src/useDismiss.ts
Dashice: This was originally called useClickOutside, but then it became useDismiss.
Reasoning: 50% of this hook concerns clicking with a pointer and 50% has to do with keyboard interactions.

packages/react-utils/src/useDismiss.ts
Dashice: These six lines check the following:

1. Whether you press an `element ` and whether your `target ` is an `Element `. This is for TypeScript happiness and will never occur IRL, unless you don't pass an Element into hook, in which case TSC will complain.
2. If you try and pass your own `pointer`, TS will ensure your configuration for this property is correct.
3. This one **should not** have a console.warn. It just ensures we don't dismiss on the wrong pointer type. E.g. dev specified only `"pen"` should dismiss and end-user uses `"touch"`.
4. This is in combination with our useInert. If whatever you clicked was inert, we do not consider that a valid dismissal; no reason to tell developers this in their logs.
5. This is the element in question itself. Of course we don't want to dismiss the element we want to remain open by pressing it itself.
6. If the button pressed was **not** left-click, if `"mouse"`, we don't want to trigger a dismissal.

Hence why there is no information provided for developers via thrown errors, or `console` entries. Disagree, or do you feel that one of these warrants providing extra information to developers?

@maeertin @adamsoderstrom

packages/react-utils/src/useDismiss.ts
```diff
@@ -0,0 +1,112 @@
...
+ * @param ref - Element reference to base events on.
+ * @param callback - Callback to execute when a `keydown` or `pointerdown` event occurs.
+ * @param options - Configurable options
+ *
+ * @example
+ * ```tsx
+ * const ref = useRef<HTMLDivElement>(null)
+ * const [isOpen, setIsOpen] = useState(false)
+ *
+ * useDismiss(ref, setIsOpen.bind(null, false), { when: isOpen })
```
maeertin, line 47: todo: I agree with you that `bind` can make it cleaner in the implementation. I do think though that it would be more clear to not use `.bind()` in the example :)

packages/react-utils/src/useDismiss.ts
maeertin: Cool, sounds good to me :)

packages/react-utils/src/useDismiss.ts
maeertin: I'm fine as it is. Thanks for the thorough explanation!

maeertin (approved):
LGTU ✨

PR #391, opened 2025-01-07
feat(react-utils): add createPolymorph helper

Dashice:


packages/react-utils/src/createPolymorph.tsx
```diff
@@ -0,0 +1,159 @@
...
+        ? P['props'] & Pick<P['overrides'], 'children'>
+        : P['props'] & P['overrides']
+      : P,
+    Tag<T, TT>,
+    O
+  >,
+) => PolymorphicElement<P, Tag<T, TT>>
+
+/**
+ * A utility function, enriching component types with polymorphism support.
```
Dashice, line 120: Consider removing explanation on `ref` behavior, as it is implied by the act of polymorphing.

packages/react-utils/src/createPolymorph.tsx
```diff
@@ -0,0 +1,159 @@
+import type { DistributiveOmit, Overwrite } from '@noaignite/types'
```
Dashice, line 1: Don't consider; just change: Add `ComplexProps` to O and change O from a string union to an Options object with keys `omit` and `overrides`

Dashice:
Closed. Replaced with #403

PR #395, opened 2025-01-15
feat(utils): add `toRootRelativePath` function

adamsoderstrom:
# Background

Paths / URLs could originate from sources that us as developers have limited control over.
For example, could an editor enter `"foo"` as input in a URL field in a CMS, if no proper validation is implemented.

`toRootRelativePath` solves this, by assuring that the path isn't [directory-relative](https://developer.mozilla.org/en-US/docs/Web/API/URL_API/Resolving_relative_references#current_directory_relative) and then may lead to a 404.


packages/utils/src/toRootRelativePath.ts
```diff
@@ -0,0 +1,30 @@
...
+ */
+export function toRootRelativePath(str: string) {
+  if (URL.canParse(str)) {
+    // The string is parseable and therefore a valid absolute URL.
+    // The string is therefore sufficient to be applied as a `href` attribute.
+    // @see: https://developer.mozilla.org/en-US/docs/Web/API/URL/canParse_static
+    return str
+  }
+
+  if (str.startsWith('/') || str.startsWith('#') || str.startsWith('?')) {
```
Dashice, line 20:
```tsx
  if (str[0] && '/#?'.includes(str[0])) {
    return str
  }
```

Performs faster generally and is more concise. I tested with 1M iterations.

packages/utils/src/toRootRelativePath.ts
```diff
@@ -0,0 +1,30 @@
+/**
+ * Transforms a given path to be root relative.
+ *
+ * @example
+ * ```typescript
+ * formatHref('foo/bar') // '/foo/bar'. Prepends leading slash, since relative paths are to be
```
Dashice, line 6: TSDocs suggests the function is named differently.

packages/utils/src/toRootRelativePath.ts
maeertin: Nice one!

Dashice (approved):

maeertin (approved):
praise: Nice little helper 🙌

PR #403, opened 2025-03-03
feat(react-utils): add createPolymorph helper

Dashice:


maeertin (approved):
Amazing work on this PR! ✨

packages/react-utils/src/createPolymorph.ts
```diff
@@ -0,0 +1,161 @@
...
+ * })
+ * ```
+ */
+export const createPolymorph = <
+  P extends Props,
+  T extends ElementType,
+  C extends Exact<Config<P, T>, C> = object,
+>(
+  render: PolymorphicRenderFunction<P, T, C>,
+) => render as PolymorphicExoticComponent<P, T, C>
```
maeertin, line 161: suggestion: Perhaps just add a console error/warning if React version is 18 or lower :)

PR #411, opened 2025-03-13
feat(react-utils): add new `useControlled` hook

maeertin:
A React hook to manage controlled versus uncontrolled state for a component.

This hook encapsulates the logic required to allow a component to work either in a controlled mode (where the parent component manages the state) or in an uncontrolled mode (where the component itself manages the state).


packages/react-utils/src/useControlled.ts
```diff
@@ -0,0 +1,106 @@
+// @see https://github.com/mui/material-ui/blob/master/packages/mui-utils/src/useControlled
+/* eslint-disable react-hooks/rules-of-hooks, react-hooks/exhaustive-deps -- Based on Mui */
+/* eslint-disable tsdoc/syntax -- Allow documentation of object params */
+
+'use client'
```
adamsoderstrom, line 5: I wonder what happens during buildtime with the `'use client'` directive, since it's not at the top-top of the file? 🤔

adamsoderstrom (commented):
So cool!

packages/react-utils/src/useControlled.ts
maeertin: Nice one, just verified that the `'use client'` directive is preserved in the `dist` folder 👍

adamsoderstrom (approved):

PR #415, opened 2025-03-15
refactor(utils): rewrite createBatchedCallback logic to use debounce

maeertin:


adamsoderstrom (approved):
Of course! Nice catch, @maeertin !

PR #420, opened 2025-03-15
fix(react-utils): fix faulty check for`isRefObject`

maeertin:


packages/react-utils/src/isRefObject.test.ts
```diff
@@ -0,0 +1,35 @@
+import { createRef } from 'react'
+import { describe, expect, it } from 'vitest'
+import { isRefObject } from './isRefObject'
+
+describe('isRefObject', () => {
+  it('returns true for a createRef object', () => {
+    const ref = createRef<HTMLDivElement>()
+    expect(isRefObject(ref)).toBe(true)
+  })
```
maeertin, line 9: Interestingly the first test is failing. `Object.isExtensible(value)` inside `isRefObject` fails the test. Care to look at it @Dashice ? :)

packages/react-utils/src/isRefObject.ts
```diff
@@ -17,4 +17,4 @@ import type { RefObject } from 'react'
  * ```
  */
 export const isRefObject = (value: unknown): value is RefObject<unknown> =>
-  isObject(value) && Object.isExtensible(value) && 'current' in value
+  isObject(value) && 'current' in value
```
adamsoderstrom, line 20: Do you have any input here, @Dashice? 😊

PR #432, opened 2025-04-04
build(docs): optimize copy files logic

maeertin:


martvdmoosdijk (approved):
Good cleanup! Should've done that myself :)

PR #433, opened 2025-04-05
feat(style-guide): add new package with eslint, prettier & ts configs

maeertin:
Add new style guide package with configurations for eslint, prettier & typescript.
This package replaces:
- `@repo/eslint-config`
- `@repo/prettier-config`
- `@repo/typescript-config`


packages/style-guide/eslint/base.js
```diff
@@ -0,0 +1,42 @@
...
+  ...typescriptConfig,
+  ...bestPracticeConfig,
+  ...possibleErrorsConfig,
+  ...es6Config,
+  ...stylisticConfig,
+  ...unicornConfig,
+  ...commentsConfig,
+  ...importConfig,
+  ...turboConfig,
+  ...vitestConfig,
```
adamsoderstrom, line 41: I'm wondering if we should avoid adding `vitest` as a default here.
If one would go with another testing library, one would have to inactivate these inherited rules, if i'm not mistaken.

maeertin: This is the "base" that the monorepo uses though. If one needs something else than `vitest` one is meant to set up their own "base" config.

packages/style-guide/README.md
```diff
@@ -16,3 +16,69 @@ npm install @noaignite/style-guide
...
+- `@noaignite/style-guide/eslint/react` - For React projects
+- `@noaignite/style-guide/eslint/base` - Base configuration
+
+### Composing Custom Rule Sets
+
+You can also compose your own configuration by importing specific rule sets:
+
+```js
+// eslint.config.js
+import baseConfig from '@noaignite/style-guide/eslint/rules/base'
```
maeertin, line 45: There is no `rules/base` file to import from :)

packages/style-guide/README.md
```diff
@@ -16,3 +16,69 @@ npm install @noaignite/style-guide
...
+
+```js
+// prettier.config.js
+import prettierConfig from '@noaignite/style-guide/prettier'
+
+export default {
+  ...prettierConfig,
+  // Your custom prettier options
+}
+```
```
maeertin, line 84: Missing `### Typescript Configuration` section

adamsoderstrom (approved):

PR #437, opened 2025-04-14
feat(docs): migrate to nextra 4

maeertin:


docs/src/app/_meta.global.ts
```diff
@@ -0,0 +1,49 @@
...
+
+export default {
+  contact: {
+    type: 'page',
+    title: 'Contact us',
+    href: 'https://www.noaignite.se/contact',
+  },
+  CONTRIBUTING: 'Contributing',
+  README: 'Getting Started',
+  // '-': { type: 'separator' },
```
adamsoderstrom, line 23: **nitpick(non-blocking):** could be removed

adamsoderstrom (approved):
Soo nice, @maeertin!
I'm a bit out of the loop by the necessities on how to migrate, besides the App router interfaces. Though, as i interpret it, the heavy lifting in this PR is more in regards to Nextra than Next.

Great initiative!

PR #457, opened 2025-04-15
docs(react-centra-checkout): add disclaimer regarding `Context.initialSelection`

adamsoderstrom:


.changeset/big-cherries-destroy.md
```diff
@@ -0,0 +1,5 @@
+---
+'@noaignite/react-centra-checkout': patch
+---
+
+fix(react-centra-checkout): add disclaimer regarding `Context.initialSelection`
```
maeertin, line 5: thought: Although nothing wrong, the changeset is not meant to be conventional commits :)

PR #466, opened 2025-04-16
feat(style-guide): export multiple prettier configs

maeertin:


adamsoderstrom (approved):
Not having to handle tailwind in the prettier config sounds very good, imo!

PR #469, opened 2025-04-17
fix(tailwind-typography): fix incorrect fontSize if clampMax true

maeertin:


packages/tailwind-typography/src/index.ts
```diff
@@ -39,7 +37,7 @@ function fluidValue(
     return `max(${minValue}${unit}, ${value})`
   }
   if (clampMax) {
-    return `min(${minValue}${unit}, ${value})`
+    return `min(${maxValue}${unit}, ${value})`
```
adamsoderstrom, line 40: Oh! Good catch!

adamsoderstrom (approved):

PR #472, opened 2025-04-18
fix(docs): add redirect from `/` to `/README`

maeertin:


adamsoderstrom:
Added some work here, to avoid the redirect, @maeertin

maeertin:
> Added some work here, to avoid the redirect, @maeertin

Ahh this is great, thanks for the help ✨

adamsoderstrom (approved):

PR #490, opened 2025-04-20
fix(style-guide): add `node.js` ESlint config 

adamsoderstrom:
# Why?

- Global variables like `__dirname` and `process.env.` are otherwise not recognized.
- The [`eslint-config-next`](https://github.com/vercel/next.js/blob/21d2a32d50164f877917c6c79dfc0e6686a5bb59/packages/eslint-config-next/index.js#L125-L128) package already adds it

### DIFF
--- a/.changeset/loud-cougars-clap.md
+++ b/.changeset/loud-cougars-clap.md
@@ -1,0 +1,5 @@
+---
+'@noaignite/style-guide': patch
+---
+
+fix(style-guide): add `node.js` ESlint config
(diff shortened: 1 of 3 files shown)


packages/style-guide/eslint/rules/next.js
```diff
@@ -10,5 +11,11 @@ export default [
       ...pluginNext.configs.recommended.rules,
       ...pluginNext.configs['core-web-vitals'].rules,
     },
+    languageOptions: {
+      globals: {
+        ...globals.browser,
```
maeertin, line 16: thought: `...globals.browser,` is already set up by `react.js`, dunno if it should be included twice.

packages/style-guide/eslint/rules/next.js
```diff
@@ -10,5 +11,11 @@ export default [
...
+        ...globals.node,
```
maeertin, line 17: thought: `...globals.node,` should perhaps be set in `javascript.js` as it could be used anywhere, not only in next evironments.

maeertin (commented):
praise: Nice one, this was actually missing. Left some thoughts on tweeks :)

PR #504, opened 2025-04-26
feat(style-guide): add new import rules and add other rules documentation

maeertin:


adamsoderstrom (approved):
Quality stuff! 🌟
Though, i'm wondering if this should be something else than a "feat"?

Definitely a nitpick, but i want to hear your point of view.

maeertin:
> i'm wondering if this should be something else than a "feat"?

As far as I understand it, a conventional commit needs to be either a fix or a feat for a change to be version bumped and published. And since this change is definitively not a fix I went with feat @adamsoderstrom :)

PR #514, opened 2025-04-30
feat(types): extend N with number[]

Dashice:
`Repeat<T, N>` type helper is a fantastic way to quickly specify a tuple of a repeated value.

```tsx
type SocialIcons = Repeat<Image, 4>;
// [Image, Image, Image, Image]
```

One limitation that often can be a painpoint for developers is the need to supply a union of tuples. Some common examples are:

- A footer navigation with link lists, but where the footer can support 1-4 of such lists due to space limitations.
- A gallery block which can render a specific set-size of images (3, 6, 9), but no number below 3, above 9 or between 3-6 and 6-9.

Intuitively, one would assume that the below should allow for this, but it doesn't.

```tsx
type Links = Repeat<Link, 1 | 4>;
// [Link]
```

TypeScript requires a singular numeric expression when it is to be used in a type iteration for length, a union of numbers does not ensure that TypeScript iterates over the context once for each union-member.

---

This change enables `N` to be a `number | number[]` and tweaks the type-helper to support recursion for each array member of lengths, achieving this desired output.

```tsx
type Links = Repeat<Link, [1, 4]>;
// [Link] | [Link, Link, Link, Link]
```

Before this change, in order to achieve this; one would need to repeat the entire repeat type expression in a union, which felt clunky.

```tsx
type Links = Repeat<Link, 1> | Repeat<Link, 4>
// [Link] | [Link, Link, Link, Link]
```


maeertin (approved):
Nice addition, looks good to me ✨

PR #525, opened 2025-05-11
fix(style-guide): avoid `declaration` and `declarationMap` in `tsconfig.next.json`

adamsoderstrom:
`declaration` and `declarationMap` was inherited from `tsconfig.base.json`, but now overwritten in `tsconfig.next.json`. 

Disabling these rules, because:
- `declaration` and `declarationMap` is not needed in a web app, since it doesn't generate types for consumption of other parties.
- `next` doesn't include them in their default `tsconfig`
- Resolves a `ts(2742)` error, suggesting a type annotation on React components (related to `@types/react@19.1.0`).


vincentboiardt:
I do not have enough history to know why it's in `tsconfig.base.json` to begin with, otherwise I think you'd have to argue why it shouldn't be removed there instead of overwritten in `tsconfig.next.json`.

maeertin (commented):
question: Why should we avoid this? Scaffolding with turbo has these two set to `true`.

PR #552, opened 2025-06-23
fix(react-utils): add stable reference of callback, but execute with …

Dashice:
Fixes an issue pointed out by @martvdmoosdijk.

```tsx
const [value, setValue] = useState(0)

useEvent('window', 'resize', () => {
  setValue(value + 1) // updates first time, then never updates again
})
```

Previous method of stabilizing callbacks inside hooks would stabilize the reference to `callback` but also its returned values. Using `useStableCallback` hook, for DRY-principle, ensures now that the `callback` never changes reference, but when `callback` is executed, it will always use the newest version of the stabilized function.


adamsoderstrom (approved):

packages/react-utils/src/useStableCallback.ts
```diff
@@ -0,0 +1,31 @@
...
+ *
+ * @example
+ * ```tsx
+ * const stableCallback = useStableCallback(() => getRandomValue())
+ * stableCallback()
+ * // always returns the latest value, but the reference to the
+ * // callback itself is stable and never changes
+ * ```
+ */
+export const useStableCallback = <A extends unknown[], T>(callback: (...args: A) => T) => {
```
maeertin, line 23: thought: Have we weighed in alternative names for this hook? For example `useMemoizedCallback`, `useStoredCallback` or similar? I found the naming of "stable" somewhat confusing. It got me wondering what it is that is unstable to start with. An unmemoized or recreated function isn't necessarily "unstable".

.changeset/better-socks-stand.md
```diff
@@ -0,0 +1,5 @@
+---
+'@noaignite/react-utils': minor
+---
+
+fix stable references of callbacks with stale values
```
maeertin, line 5: This changeset might need two separate entries. The version change describes a `minor` whilst the description describes a patch. This should ideally be:

Changeset 1
```
---
'@noaignite/react-utils': minor
---

Add new `useStableCallback` hook
```

Changeset 2
```
---
'@noaignite/react-utils': patch
---

Fix stable references of callbacks with stale values
```

I have not tried this myself but perhaps that this could be a single changeset like below:
```
---
'@noaignite/react-utils': minor
---

Add new `useStableCallback` hook
Fix stable references of callbacks with stale values
```

maeertin (changes requested):
Great PR 🔥 Left some todos & given the small size of the new added hook it should probably come with a test.

packages/react-utils/src/useStableCallback.ts
adamsoderstrom: Or `useReferencedCallback`?

When i talked to Dash about it, he showed me that the naming `useStableCallback` appears to be a convention:
- https://github.com/mariusbrataas/useStableCallback
- https://foundations.significa.co/hooks/use-stable-callback
- https://gist.github.com/Shrugsy/5e898173c965e7642db8927636bedf7a

PR #553, opened 2025-06-23
fix(use-sticky): inconsistent fallback target

Dashice:
Previously, element used to determine stickiness was the `document.body`. This worked on sites whose `body` was styled in an opinionated manner. This has been replaced with a representation of `Viewport` instead, meaning the dimensions of your browser window over arbitrary `body` dimensions, which can overflow / underflow the `Viewport`.

### DIFF
--- a/packages/react-utils/src/useSticky.test.ts
+++ b/packages/react-utils/src/useSticky.test.ts
@@ -1,3 +1,4 @@
+import { capitalize } from '@noaignite/utils'
 import { act, renderHook, waitFor } from '@testing-library/react'
 import { afterEach, beforeAll, describe, expect, it, vi } from 'vitest'
 import { IntersectionObserverMock } from '../test'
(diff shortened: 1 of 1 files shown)


packages/react-utils/src/useSticky.ts
```diff
@@ -132,11 +132,17 @@ export const useSticky = (
...
         const element = ref.current
-        if (!(parent instanceof HTMLElement) || !(element instanceof HTMLElement)) return
+        if (!(element instanceof HTMLElement)) return
+
+        const pRect = parent?.getBoundingClientRect() ?? {
+          top: 0,
+          left: 0,
+          bottom: window.innerHeight,
+          right: window.innerWidth,
+        }
```
maeertin, line 144: suggestion: You could do `new DOMRect(x, y, width, height)` instead of object for stricter typings :) https://developer.mozilla.org/en-US/docs/Web/API/DOMRect/DOMRect#syntax

maeertin (approved):
LGTM

PR #585, opened 2025-09-04
fix(react-centra-checkout): unable to call method on base class

adamsoderstrom:
Tried to extend the `ApiClient` class, while i receieved a `TS2340`
error from TypeScript. Defining the `request` property as a method (and
not a function) resolves that error.

Related sources for solution:
- https://stackoverflow.com/questions/55422250/typescript-cannot-call-super-method-on-child-class
- https://codereviewvideos.com/how-i-fixed-ts2340-only-public-and-protected-methods-of-the-base-class-are-accessible-via-the-super-keyword/

### DIFF
--- a/.changeset/lazy-areas-allow.md
+++ b/.changeset/lazy-areas-allow.md
@@ -1,0 +1,5 @@
+---
+'@noaignite/react-centra-checkout': patch
+---
+
+Resolve TS2340 error while trying to call `super.request` from sub class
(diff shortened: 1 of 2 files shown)


adamsoderstrom:
Some more context.

```typescript
export class ExtendedApiClient extends ApiClient {
  request(...params: Parameters<InstanceType<typeof ApiClient>['request']>) {
    const [method, endpoint, payload] = params
    let req: ApiRequest = { method, endpoint, payload }

    this.middlewares.forEach(async (middleware) => {
      const updated = await middleware(req)
      if (updated) req = updated
    })

    return super.request(req.method, req.endpoint, req.payload) // <--- This will result in a TS2340 error, since the `request` property inside `ApiClient` is not considered a method by TypeScript, but instead a function attached to a property.
  }
```

PR #598, opened 2025-10-01
feat(react-utils): add the observer as the secondary callback arg

maeertin:
Closer to the native `IntersectionObserver` API callback. Reaching for the internal observer can be helpful.


Dashice (approved):
Makes sense without being some arbitrary feature creep. Good stuff.

PR #766, opened 2026-05-15
ops(ci): fix slack release notification payload parsing

maeertin:


adamsoderstrom (approved):
Without looking through the details of `buildPackageDocs.ts`, i do think that the intent of moving the logic to a separate file is good!

maeertin:
> Without looking through the details of `buildPackageDocs.ts`, i do think that the intent of moving the logic to a separate file is good!

I removed this change, it was originally introduced to fix an eslint error but that is no longer the case 👍

PR #801, opened 2026-07-02
feat: introduce `@noaignite/next-klaviyo` package

adamsoderstrom:


patrickchan-noaignite (approved):

packages/next-klaviyo/src/types/global.d.ts
```diff
@@ -0,0 +1,11 @@
...
+  interface Window {
+    klaviyo?: {
+      identify?: (payload: unknown) => void
+      push?: (args: unknown[]) => void
+      track?: (eventName: string, payload?: unknown) => void
+    }
+  }
+}
+
+export {}
```
maeertin, line 11: question: What does this do? Is it to disable automatic exports? Perhaps nice to add a simple explanation :)
```ts
// Disable automatic export
export {}
```

packages/next-klaviyo/src/index.ts
```diff
@@ -0,0 +1,5 @@
+export { klaviyo } from './klaviyo'
+export { KlaviyoObject } from './KlaviyoObject'
+export { KlaviyoScript } from './KlaviyoScript'
+export type { KlaviyoScriptProps } from './KlaviyoScript'
```
maeertin, line 4: question: Why not just `export *`?

packages/next-klaviyo/package.json
```diff
@@ -0,0 +1,53 @@
...
+  "author": "NoA Ignite",
+  "sideEffects": false,
+  "type": "module",
+  "exports": {
+    ".": {
+      "types": "./src/index.ts",
+      "default": "./src/index.ts"
+    }
+  },
+  "types": "./src/index.ts",
```
maeertin, line 29: question: is this needed? I was in the understanding that the `exports` field is enough :)

packages/next-klaviyo/package.json
```diff
@@ -0,0 +1,53 @@
...
+  "files": [
+    "src",
+    "dist"
+  ],
```
maeertin, line 33: question: Do we need/want to publish both `src` & `dist`?

packages/next-klaviyo/package.json
```diff
@@ -0,0 +1,53 @@
...
+  "license": "MIT",
+  "author": "NoA Ignite",
+  "sideEffects": false,
+  "type": "module",
+  "exports": {
+    ".": {
+      "types": "./src/index.ts",
+      "default": "./src/index.ts"
+    }
+  },
```
maeertin, line 28: question: What's the build script for if we point to `src` and not `dist`? Probably some faulty wiring here :)

packages/next-klaviyo/package.json
```diff
@@ -0,0 +1,53 @@
...
+  "devDependencies": {
+    "@noaignite/style-guide": "workspace:*",
+    "@types/react": "^19.2.17"
+  },
+  "peerDependencies": {
+    "next": "^14.0.0 || ^15.0.0 || ^16.0.0",
+    "react": "^18.0.0 || ^19.0.0"
+  },
+  "engines": {
+    "node": ">=20.0.0"
```
maeertin, line 48: todo(non-blocking): We should probably raise this to at least version 22 which is actively maintained :)

maeertin (commented):
Great work @adamsoderstrom 🥳 Left you some questions/todos. Resolve them and merge when you feel ready.

packages/next-klaviyo/src/types/global.d.ts
adamsoderstrom: Added a comment explaining this — it marks the file as a module so `declare global` augments correctly instead of becoming a local declaration.

packages/next-klaviyo/src/index.ts
adamsoderstrom: Switched to `export *` for the components. Kept a named export for `klaviyo` so we don’t re-export the internal `notifyKlaviyoLoaded` helper.

packages/next-klaviyo/package.json
adamsoderstrom: Removed — `exports` is enough, matching the other packages.

packages/next-klaviyo/package.json
adamsoderstrom: Agreed — now publishing only `dist`.

packages/next-klaviyo/package.json
adamsoderstrom: Fixed — `exports` now points at the built `dist` output (ESM + CJS), same pattern as the other packages.

packages/next-klaviyo/package.json
adamsoderstrom: Bumped to `>=22.0.0`.

PR #814, opened 2026-08-14
fix(react-utils/createPolymorph): allow components to set `displayName`

maeertin:
`createPolymorph` returns the original render function at runtime, so React components created by it can already have `displayName` assigned. Its TypeScript return type only described the callable signature, however, valid assignments such as `Button.displayName = 'Button'` to fail type checking. This change exposes the optional property in the public type and adds regression coverage.

### DIFF
--- a/packages/react-utils/src/createPolymorph.ts
+++ b/packages/react-utils/src/createPolymorph.ts
@@ -188,9 +188,10 @@ export type PolymorphicRenderFunction<
   P extends Props,
   T extends ElementType = ElementType,
   C extends Exact<Config<P, T>, C> = object,
-> = (
-  props: PolymorphicProps<P, T, C>,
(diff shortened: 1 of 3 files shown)


packages/react-utils/src/createPolymorph.ts
Dashice, line 269: https://www.npmjs.com/package/react-polymorphic

Here, we assign a default value of "PolymorphicComponent" if `displayName` is not set, which helps differentiate a standard unnamed component from a polymorphic one, thoughts?

Dashice (approved):
Approved with one optional comment.

packages/react-utils/src/createPolymorph.ts
maeertin: Cheers, now updated 👍

PR #833, opened 2026-09-07
fix(style-guide): enforce public test imports

maeertin:
* Require tests to import public APIs through supported package exports.
* Add missing `isRefObject` export from the package root.


adamsoderstrom (approved):
Nice one! ⭐

PR #842, opened 2026-09-23
chore(deps): bump test deps and reuse JSDOM env

maeertin:
## Speed Up Vitest JSDOM Tests

Use `vmThreads` to reuse JSDOM environments while preserving per-file isolation. Update test setup and assertions for VM compatibility.

Full uncached workspace unit-test suite: **15.439s → 8.296s**, about **46% faster**.


packages/react-centra-checkout/src/Context.test.tsx
```diff
@@ -252,7 +245,7 @@ describe('CentraProvider', () => {
             {
               token: 'foo',
               action: 'redirect',
-              url: 'https://example.com/checkout',
+              url: '#checkout',
```
maeertin: Use hash-only navigation in checkout test fixtures because JSDOM cannot perform cross-document navigation. This preserves redirect coverage while preventing noisy “Not implemented: navigation” test output.

packages/react-centra-checkout/src/PaymentEmbed.test.tsx
```diff
@@ -17,7 +17,7 @@ function CentraProviderWrapper(props: { children: React.ReactNode }) {
       apiUrl={CENTRA_API_URL}
       paymentFailedPage=""
       paymentReturnPage=""
-      receiptPage=""
+      receiptPage="#"
```
maeertin: Use hash-only navigation in checkout test fixtures because JSDOM cannot perform cross-document navigation. This preserves redirect coverage while preventing noisy “Not implemented: navigation” test output.

packages/utils/src/URLSearchParamsFromObject.test.ts
```diff
@@ -12,7 +12,7 @@ describe('URLSearchParamsFromObject', () => {
     const params = URLSearchParamsFromObject(object)
 
     expect(params.get('name')).toEqual(object.name)
-    expect(params.getAll('hobbies')).toStrictEqual(object.hobbies)
+    expect(params.getAll('hobbies')).toEqual(object.hobbies)
```
adamsoderstrom: Why do we need to change from `toEqual` here?

oelburk (approved):

